PR #226, opened 2013-07-04
Use Capistrano's current_path.

rroblak:
Use Capistrano's current_path instead of release_path to allow the Cap
task to be run independent of when the release directory is created.
The release directory name is determined at Cap load time via Time.now;
however, if the release directory is not being created during this
Cap run the Airbrake Cap task tries to cd into a non-existent directory
and errors out.  Better to use current_path, which is always
correct.


kyrylo:
Hi! Does this also work when you deploy for the very first time?

rroblak:
@kyrylo Not sure as I don't use Cap any more.

kyrylo:
Thanks for the answer. I'm going to close this for now, since there were no other reports for this issue. Let me know if it's still an issue in case someone stumbles upon it.

PR #265, opened 2013-12-30
Support nested exceptions

dvdplm:
Support the nested exceptions feature of Ruby 2.1 (Exception#cause) in a backwards compatible way (i.e. without breaking anything).
Implies a pretty big refactoring of how the Notice class renders exception(s) and tries to untangle the various custom options to notice creation from the actual exceptions.


Jae:
I would also love to see this happening!

kyrylo:
This will be featured in our next major release (soon).

PR #272, opened 2014-03-06
Remove code duplication for Rack middleware

omerisimo:
Browsing through the code in an attempt to understand the logic for sending notifications I found that *Airbrake::Rack* and *Airbrake::Rails::Middleware* classes had a lot of common logic.
This made the code hard to follow and understand the differences between the two implementations.

This pull request makes the code of the two classes DRY :) [Don't Repeat Yourself]
All tests are passing, and I have performed some manual testing to make sure the full integration works well in both Rails 3.2 and Sinatra.

### DIFF
--- a/lib/airbrake/rack.rb
+++ b/lib/airbrake/rack.rb
@@ -25,20 +25,6 @@ module Airbrake
       Airbrake.configuration.framework = "Rack: #{::Rack.release}"
     end
 
-    def ignored_user_agent?(env)
-      true if Airbrake.
-        configuration.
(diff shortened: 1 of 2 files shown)


kyrylo:
Hi! I appreciate your work, but this looks outdated now :(

I would go for composition instead of inheritance, since this line `class Middleware < Airbrake::Rack` doesn't seem to be logical for me.

I couldn't find an explanation of why some methods are private and others are protected.

If you're still interested, feel free to open a new PR.

PR #290, opened 2014-05-09
Add ability to define whitelist filters for parameters

omerisimo:
I found that filtering parameters from being sent to Airbrake using the :params_filters attribute is something that I constantly need to update and think about. (In my company we need to make sure no personal customer information is shared with 3rd party services)

Since there is only a small number of parameters I actually care about when examining an Airbrake notification, I think that the ability to whitelist the parameters I am interested in, is a simpler, safer approach.

By setting the :params_whitelist_filters list, all other parameters will be filtered (and their content replaced).
If a parameter is both whitelisted and "blacklisted" it will be filtered.


shifi:
@omerisimo I've changed the code a little bit to match new changes and few other stuff.
Thank you!

PR #295, opened 2014-05-19
Verbose output when notice not sent

alif:
Write verbose output when configuration prevents sending notices to Airbrake.


benoittgt:
Is it possible to disable it when running test with [asynchronous airbrake](https://github.com/airbrake/airbrake/wiki/Using-Airbrake-asynchronously) ? 

image

I've try to do : 
```

PR #314, opened 2014-08-08
ignore empty strings in gempath.

mauriciomenegaz:
This is a fix for an issue that happens when running rails with spring (see https://github.com/rails/spring/issues/310). For some reason Gem.path contains an empty string, which causes the default backtrace filter to behave incorrectly. With this change, empty strings in Gem.path are ignored.

### DIFF
--- a/lib/airbrake/configuration.rb
+++ b/lib/airbrake/configuration.rb
@@ -131,7 +131,7 @@ module Airbrake
       lambda { |line| line.gsub(/^\.\//, "") },
       lambda { |line|
         if defined?(Gem)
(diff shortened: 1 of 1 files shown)


shifi:
@mauriciomenegaz this issue is fixed with [this PR](https://github.com/airbrake/airbrake/commit/038679973f5d2ff6dd7a9576ad5b96f73a953d55).
So sorry I didn't notice your PR first (this approach looks much better). If you create a new PR including this change, I will merge it. Just make sure all the tests are passing.
Thank you!

PR #321, opened 2014-08-25
Adds airbrake_stages and airbrake_role configurations.

seankibler:
In an environment I'm working in we have multiple app servers and use multiple stages (dev, staging,  production). We really only want Airbrake to run notifications in when we deploy to production and only on one server (not for every app server) per deploy. I imagine other people would benefit from these capabilities as well.

I believe the way I have written these changes basic usage of Airbrake Capistrano v3 integration will operate exactly as it does now. In other words it is not necessary to configure any of these options, it will just work with the defaults. I have attempted to merely add further ability to limit when and how much Airbrake will be applied :)

I'll happily contribute further by updating the relevant documentation if my pull request is accepted.

### DIFF
--- a/lib/airbrake/tasks/airbrake.cap
+++ b/lib/airbrake/tasks/airbrake.cap
@@ -11,7 +11,7 @@ namespace :airbrake do
       - Run remotely so we use remote API keys, environment, etc.
   DESC
   task :deploy do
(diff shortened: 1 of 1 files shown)


kyrylo:
This branch is a bit outdated now and should be rebased. According to the lack of feedback, it seems like not many people are interested in this. Thanks.

Issue #313, opened 2014-08-04
PostgreSQL Connection issues with Rails 4.1.2


reconbot:
We've started a new app with Rails 4.2.1, added airbrake 4.0.0 to it and then started having issues connecting to the database. After a few page loads we started running out of available database connections for our connection pool. Tracking it down further, it seemed a broken asset link in template was causing exceptions to be thrown in rails, and each one of those connections connected to the db but never released the connection. A found a similar bug with [rollbar](https://github.com/rollbar/rollbar-gem/issues/70).

AteqEjaz:
Hi,

Airbrake is just showing me this on Dashboard

PG::ConnectionBad: could not connect to server: Connection refused Is the server running on host "localhost" (127.0.0.1) and accepting TCP/IP connections on port 5433? 

Any suggestions? 

Using Rails
gem 'rails', github: 'rails/rails', branch: '4-0-stable' with Ruby 2.x

Thanks.

shifi:
@teknull I just released new version which should include fix for this issue.
I will close it but feel free to reopen it if you still have see any related issues.

shifi:
@AteqEjaz please update Airbrake gem version to 4.2.0. Thank you!

teknull:
@shifi Is this new gem compatible with Rails 3.2.21?  I'm using Airbrake in a legacy app and need support for it.  Also on Ruby 1.9.3, FYI

shifi:
@teknull the last tested Rails 3.x.x version is `3.2.13`.
I just tested against 3.2.21 locally and all tests passed so I guess it should be compatible with 3.2.21.
If you notice any issues, please let me know.
Thanks

teknull:
@shifi Thanks, I'll upgrade the gem in development (although I could never get it to throw errors in development and AirBrake's support was non-responsive) and give it a test.  I'll probably throw it into a production environment (staging) and let you know if I see any problems.  Thanks for the gem bump and hard work!

PR #331, opened 2014-09-18
Fixes airbrake/airbrake#313

tonyarkles:
Accessing the `current_user` method on ApplicationController can cause
a database access when an exception is thrown. This database access
acquires a connection from the Rails connection_pool, but there's no
clean-up to return that connection back to the pool. Eventually, the
connection pool would be empty and the application fails.


toymachiner62:
Has this been released yet? If so, what version of airbrake? I'm also hitting a connection timeout issue with mysql with my latest code changes and I did add some code that references devise' `current_user` in my application controller.

PR #333, opened 2014-09-20
Remove indents created by generator automatically

kirikiriyamama:
So far, the lines inserted by generator have been indented like the
following lines. I think it is wrong.

```
        require './config/boot'
        require 'airbrake/capistrano'
```


shifi:
@kirikiriyamama there has been a few changes related to Capistrano lately.
We removed indents as part of [this commit](https://github.com/airbrake/airbrake/commit/53a7c0a6ed97ff1010c7daa57bba3de87a6cd0d6).
Thank you!

PR #334, opened 2014-09-23
Refactor #airbrake_current_user

dvdplm:
Fix for https://github.com/airbrake/airbrake/issues/313 based on the (excellent) work by @tonyarkles and his PR https://github.com/airbrake/airbrake/pull/331
- ensure we return the DB connection  
- split the DB interaction from populating the hash

The reason I prefer doing it this way is that it isolates the problematic spot into a single short method.

### DIFF
--- a/lib/airbrake/rails/controller_methods.rb
+++ b/lib/airbrake/rails/controller_methods.rb
@@ -103,7 +103,9 @@ module Airbrake
           nil
         end
       ensure
-        if(defined?(ActiveRecord) && ActiveRecord::Base.respond_to?(:connection_pool))
+        # The Airbrake middleware is first in the chain, before ActiveRecord::ConnectionAdapters::ConnectionManagement
+        # kicks in to do its thing. This can cause the connection pool to run out of connections.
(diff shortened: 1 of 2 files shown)


jbaudanza:
I just traced a connection pool leak in my app down to `airbrake_current_user`. It might be time to do another gem release so this fix gets pushed out to your users. I imagine it's affecting a lot of people.

teknull:
I agree.  I think this was addressed in a previous PM/merge but I don't believe the gem has been released to rubygems/etc.  They are still on 4.1.0 from 9/4/14.  This definitely needs to be released as I've had this issue too.

jbaudanza:
This feature should probably removed, or at least make it opt-in. The default implementation of `airbrake_current_user` should return `{}`. This is good enough for the vast majority of applications with a user_id in the session.

You have no idea what `current_user` really does or what its side effects are. It's also very unintuitive that this middleware would be doing this and I don't see this behavior documented anywhere.

For anyone else having this prolem, my workaround was do override airbrake_current_user in my ApplicationController:

```
  def airbrake_current_user
    {}
  end
```

PR #339, opened 2014-10-03
Expose configuration settings to set api version

dvdplm:
This PR exposes a way for users to pick the Airbrake API version to use to submit errors. The default is maintained to the old 2.4 XML-based api. This default will change to use v3 in a near future.


kyrylo:
1 year anniversary of this issue! =) This branch has conflicts. We will have something very special for the new API.

PR #349, opened 2014-11-15
Insert UserInformer middleware before Rack::Runtime 

phuongnd08:
Since Rack::Lock may not be available for app with websocket-rails


shifi:
@phuongnd08 which Rails version you're using?
Can you provide me with steps to reproduce this? Thanks

shifi:
Thanks @phuongnd08

PR #353, opened 2014-12-05
filter senstive info

eadz:
Adding HTTP_AUTHORIZATION to rack filters,

also SENSITIVE_ENV_VARS currently has secret and password, I've added auth and token here as well.

### DIFF
--- a/lib/airbrake/utils/rack_filters.rb
+++ b/lib/airbrake/utils/rack_filters.rb
@@ -2,6 +2,7 @@ module Airbrake
   SENSITIVE_RACK_VARS = %w(
       HTTP_X_CSRF_TOKEN
       HTTP_COOKIE
+      HTTP_AUTHORIZATION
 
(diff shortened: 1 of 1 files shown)


shifi:
@eadz if we add that, this will affect all Airbrake users since this part is not configurable.
We are ok to add only `HTTP_AUTHORIZATION` for now until we make this configurable.

eadz:
Hi @shifi,

Since I did this PR, It is my experience that sensitive info always leaks out. I almost think a whitelist based approach would be better...

shifi:
@eadz I will add only `HTTP_AUTHORIZATION` for now. We implemented whitelist based approach so you can use that if you like. Currently it's on master and it will be out with next minor release.
In the mean time, if you use edge version of the gem, you can use whitelist approach like this:
`config.params_whitelist_filters << "some_key"` or
`config.params_whitelist_filters = [:foo, 'bar', /something_else/]`

PR #354, opened 2014-12-11
Assumption of existence of :app role may not be valid

wojobucco:
I was setting up a new deployment in cap3 after migrating from cap2.  I did not have an :app role defined for my stages, only :web and :db.  This caused the airbrake notification not to happen.  I recommend that this be changed to :all role to keep that from happening.

Looking at the file history, this is actually how it used to be defined.  Was this changed to :app for any reason?


shifi:
You're right @wojobucco 
Thank you!

PR #355, opened 2014-12-12
Update sidekiq.rb

seuros:
instead of testing Sidekiq version, let use gem to select the correct version

### DIFF
--- a/lib/airbrake/sidekiq.rb
+++ b/lib/airbrake/sidekiq.rb
@@ -1,9 +1,8 @@
+gem 'sidekiq', '~> 3.0'
 require 'sidekiq'
 
-if ::Sidekiq::VERSION >= '3'
-  ::Sidekiq.configure_server do |config|
-    config.error_handlers << lambda do |exception, context|
-      Airbrake.notify_or_ignore(exception, :parameters => context)
-    end
+Sidekiq.configure_server do |config|
+  config.error_handlers << ->(exception, context)
+    Airbrake.notify_or_ignore(exception, :parameters => context)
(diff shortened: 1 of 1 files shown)


sheerun:
Is it RubyGems feature?

seuros:
no, Ruby.
rubygems/core_ext/kernel_gem.rb

lib/airbrake/sidekiq.rb
```diff
@@ -1,9 +1,8 @@
+gem 'sidekiq', '~> 3.0'
 require 'sidekiq'
 
-if ::Sidekiq::VERSION >= '3'
-  ::Sidekiq.configure_server do |config|
-    config.error_handlers << lambda do |exception, context|
-      Airbrake.notify_or_ignore(exception, :parameters => context)
-    end
+Sidekiq.configure_server do |config|
+  config.error_handlers << ->(exception, context)
```
shifi: @seuros did you missed `do` at the end of the line?
`  config.error_handlers << ->(exception, context) do`
Without it I get syntax error.

shifi: Also, I think lambda literal syntax is better when you have single line block, but I guess that is question of personal taste :)

seuros: I will revert to lambda syntax then. Thank you

shifi: Thanks

shifi: @seuros I changed that for you. Thank you!

PR #363, opened 2015-01-18
fix bug: rails request.fullpath behavior change in 4.1.6 -- no longer has leading /

halida:
According to the issue: https://github.com/rails/rails/issues/16926

When rails version > 4.1.6, request.fullpath no longer has leading /, which will break Airbrake::Rails::ControllerMethods#airbrake_request_url, so make the code support those two cases.


shifi:
thank you @halida 
nice catch!

PR #376, opened 2015-04-02
This should fix #369

### LINKED ISSUE #369: RuntimeError (can't modify frozen Array): lib/airbrake/utils/params_cleaner.rb:111:in `collect!'
I'm using a simple static page renderer gem called high_voltage. It triggers a call to Rack::Sendfile#call, which passes in the entire env. The rack version array ([1, 0]) is frozen, but airbrake tries to modify it using Array#collect!. Details below. If I add "&& ! data.frozen?" to the condition on lines 111 and 115, then it works as expected, but I'm not sure that's a proper fix.

In my pry session:

    [3] pry(#<Airbrake::Utils::ParamsCleaner>)> opts[:to_clean][:cgi_data]["rack.version"]
    => [1, 0]
    [4] pry(#<Airbrake::Utils::ParamsCleaner>)> opts[:to_clean][:cgi_data]["rack.version"].frozen?   
    => true

Here's the method definition: 

     98: def clean_unserializable_data(data, stack = [])
     99:   binding.pry if data.frozen?
    100:   return "[possible infinite recursion halted]" if stack.any?{|item| item == data.object_id }
    101:   if data.is_a?(String)
    102:     data
    103:   elsif data.is_a?(Hash)
    104:     data.inject({}) do |result, (key, value)|
    105:       result.merge!(key => clean_unserializable_data(value, stack + [data.object_id]))
    106:     end
    107:   elsif data.respond_to?(:to_hash)
    108:     data.to_hash.inject({}) do |result, (key, value)|
    109:       result.merge!(key => clean_unserializable_data(value, stack + [data.object_id]))
    110:     end
    111:   elsif data.respond_to?(:collect!)
    112:     data.collect! do |value|
    113:       clean_unserializable_data(value, stack + [data.object_id])
    114:     end
    115:   elsif data.respond_to?(:to_ary)
    116:     data.to_ary.collect! do |value|
    117:       clean_unserializable_data(value, stack + [data.object_id])
    118:     end
    119:   elsif data.respond_to?(:to_s)
    120:     data.nil? ? nil : data.to_s
    121:   end
    122: end

MGPalmer:
* Don't modify data in place with collect!
* Don't try to "collect" over IO objects

### DIFF
--- a/lib/airbrake/utils/params_cleaner.rb
+++ b/lib/airbrake/utils/params_cleaner.rb
@@ -106,12 +106,12 @@ module Airbrake
             data.to_hash.inject({}) do |result, (key, value)|
               result.merge!(key => clean_unserializable_data(value, stack + [data.object_id]))
             end
-          elsif data.respond_to?(:collect!)
-            data.collect! do |value|
+          elsif data.respond_to?(:collect) and not data.is_a?(IO)
+            data = data.collect do |value|
               clean_unserializable_data(value, stack + [data.object_id])
             end
           elsif data.respond_to?(:to_ary)
-            data.to_ary.collect! do |value|
+            data = data.to_ary.collect do |value|
               clean_unserializable_data(value, stack + [data.object_id])
             end
           elsif data.respond_to?(:to_s)


shifi:
@MGPalmer thank you!

MGPalmer:
\o/ Thanks !

PR #382, opened 2015-05-20
Use proper capistrano hook depending on capistrano version

### LINKED ISSUE #324: Generator should generate correct capistrano integration
At the moment, the generator probably assumes capistrano v2 and always adds this integration. The generator should check the capistrano version and install the correct integration.

for cap v2:
```ruby
require './config/boot'
require 'airbrake/capistrano'
```

for cap v3:
```ruby
require 'airbrake/capistrano3'
after :"deploy:restart", "airbrake:deploy"
```

shifi:
Fixes #324

### DIFF
--- a/lib/rails/generators/airbrake/airbrake_generator.rb
+++ b/lib/rails/generators/airbrake/airbrake_generator.rb
@@ -25,7 +25,7 @@ class AirbrakeGenerator < Rails::Generators::Base
   def install
     ensure_api_key_was_configured
     ensure_plugin_is_not_present
-    append_capistrano_hook
+    append_capistrano_hook if capistrano_present?
     generate_initializer unless api_key_configured?
     determine_api_key if heroku?
     test_airbrake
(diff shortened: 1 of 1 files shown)


dvdplm:
Are we sure we even want to rewrite peoples deploy scripts like this? Is this common practise for generators? :/

vmihailenco:
Not sure I follow, but why not? Looks like we use capistrano hook to notify Airbrake about deploy. Looks okay.

shifi:
We just append capistrano hook at the end of deploy script. We did it like
that before, but we always used hook for capistrano version 2

lib/rails/generators/airbrake/airbrake_generator.rb
```diff
@@ -47,18 +47,42 @@ def ensure_plugin_is_not_present
...
+    !Gem.loaded_specs['capistrano'].nil? &&
+      File.exists?('config/deploy.rb') &&
+      File.exists?('Capfile')
+  end
 
-        require './config/boot'
-        require 'airbrake/capistrano'
-      HOOK
+  def append_capistrano_hook
+    if capistrano_version < Gem::Version.new('3')
```
phumpal: cap3 is now 2 years old. How long should we be concerned with supporting v2?

phumpal:
I tend to agree with @dvdplm, however, in this instance we improve usability so LGTM.

PR #384, opened 2015-05-28
Improves documentation for notify method

biske:
I wanted to pass additional context about errors and tried to search documentation.
Looking at Airbrake#notify method seems like it doesn't accept parameters:

https://github.com/airbrake/airbrake/blob/master/lib/airbrake.rb#L130

Comments above don't mention parameters hash.
But digging deeper I found:

https://github.com/airbrake/airbrake/blob/master/lib/airbrake/notice.rb#L111

Indeed it works. This pull request is attempt to improve `notify` method documentation.


shifi:
Looks good. Thanks!

PR #385, opened 2015-05-29
Support Rails 4 without ActiveRecord

arrtchiu:
In some Rails 4 configurations, `ActiveRecord` is defined, but `ActiveRecord::Base` is not.


shifi:
Looks great. Thanks!

Issue #347, opened 2014-11-13
Airbrake not compatible with better_errors


teknull:
I have a Rails 3.2.20 app in which I use `better_errors` and `binding_of_caller` in my development gemset.

My goal is to try and have AirBrake fire notifications when in the development environment.  After going back and forth with support and experimenting on my own it seems that `better_errors` does not allow `airbrake` to fire an exception and notify.

Here is what my Gemfile looks like

```
source 'https://rubygems.org'

# Bundle edge Rails instead:
gem 'rails',  '~> 3.2.19'
gem 'pg', "~> 0.14.0"
gem 'devise', "~> 2.1.2"
gem "combined_time_select", "~> 0.0.1"
gem "will_paginate", "~> 3.0.3"
gem 'whenever', "~> 0.7.3", :require => false
gem "gritter", "1.0.2"
gem "rails_admin", "~> 0.3.0"
gem "twitter-bootstrap-rails", "~> 2.2.1"
gem "will_paginate-bootstrap", "~> 0.2.2"
gem "prawn"
gem 'twilio-ruby', '~> 3.11.6'
gem 'carrierwave', '~> 0.10.0'
gem 'geocoder', '~> 1.2.5'
gem 'attribute_normalizer', '~> 1.2.0'
gem 'chartkick', '~> 1.3.2'
gem 'groupdate'
gem 'airbrake'




group :development do
  gem "annotate", "~> 2.5.0"
  gem "better_errors", "~> 0.3.2"
  gem "binding_of_caller"
  gem 'thin', "~> 1.4.1"
  gem 'brakeman', :require => false
end

gem "gmaps4rails", "~> 2.0.0.pre"
gem "taip_parser", "~> 0.0.1"
gem "foreman", "~> 0.60.0"

# Gems used only for assets and not required
# in production environments by default.
group :assets do
  gem 'sass-rails',   '~> 3.2.3'
  gem 'coffee-rails', '~> 3.2.1'
  gem 'therubyracer', '~> 0.12.1', :platforms => :ruby
  gem 'uglifier', '>= 1.0.3'
  gem 'jquery-datatables-rails'
end

gem "less-rails" #Sprockets (what Rails 3.1 uses for its asset pipeline) supports LESS
gem 'jquery-rails'

# Deploy with Capistrano
gem 'capistrano'
```

Here's my `airbrake.rb` initializer:
```
Airbrake.configure do |config|
  config.api_key = 'blahblahblah'
  config.development_environments = []
[...]

PR #387, opened 2015-06-03
Support for better errors

shifi:
Fixes: #347

### DIFF
--- a/lib/airbrake/better_errors.rb
+++ b/lib/airbrake/better_errors.rb
@@ -1,0 +1,38 @@
+require 'better_errors'
+
+module BetterErrors
+  class Middleware
+    alias_method :show_error_page_original, :show_error_page
+
+    private
+
+    def show_error_page(env, exception=nil)
+      env['airbrake.error_id'] = notify_airbrake(env, exception)
+      show_error_page_original(env, exception)
+    end
+
+    def notify_airbrake(env, exception)
+      unless ignored_user_agent? env
+        Airbrake.notify_or_ignore(exception, request_data(env))
+      end
+    end
+
+    def ignored_user_agent?(env)
+      true if Airbrake.configuration
+        .ignore_user_agent.flatten
+        .any? { |ua| ua === env['HTTP_USER_AGENT'] }
+    end
+
+    def request_data(env)
+      if controller(env).respond_to?(:airbrake_request_data)
+        controller(env).airbrake_request_data
+      else
+        {:rack_env => env}
+      end
+    end
+
+    def controller(env)
+      env["action_controller.instance"]
+    end
+  end
+end
--- a/lib/airbrake/better_errors.rb
+++ b/lib/airbrake/better_errors.rb
@@ -1,10 +1,10 @@
 require 'better_errors'
 
 module BetterErrors
-  # Better Errors' error handling middleware with Airbrake support. Including
-  # this in your middleware stack will report uncatched exceptions to Airbrake
-  # and show a Better Errors error page for exceptions raised below this
-  # middleware.
+  # When using Better Errors' error handling middleware, this code will
+  # first report uncatched exceptions to Airbrake and then proceed with
+  # original method call, which will show a Better Errors error page for
+  # exceptions raised below this middleware.
   #
   class Middleware
     alias_method :show_error_page_original, :show_error_page
--- a/lib/airbrake/better_errors.rb
+++ b/lib/airbrake/better_errors.rb
@@ -1,6 +1,11 @@
 require 'better_errors'
 
 module BetterErrors
+  # Better Errors' error handling middleware with Airbrake support. Including
+  # this in your middleware stack will report uncatched exceptions to Airbrake
+  # and show a Better Errors error page for exceptions raised below this
+  # middleware.
+  #
   class Middleware
     alias_method :show_error_page_original, :show_error_page
 
@@ -27,8 +32,8 @@ module BetterErrors
       if controller(env).respond_to?(:airbrake_request_data)
         controller(env).airbrake_request_data
       else
-        {:rack_env => env}
       end
+        {:rack_env => env}
     end
 
     def controller(env)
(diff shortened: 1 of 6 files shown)


lib/airbrake/better_errors.rb
```diff
@@ -0,0 +1,38 @@
+require 'better_errors'
```
vmihailenco: Please add a comment explaining the purpose of this file. Something like:

"""
BetterErrors module defines middleware that reports uncatched exceptions to Airbrake and displays default error page.
"""

and add whatever details I don't know.

lib/airbrake/better_errors.rb
```diff
@@ -0,0 +1,38 @@
...
+      show_error_page_original(env, exception)
+    end
+
+    def notify_airbrake(env, exception)
+      unless ignored_user_agent? env
+        Airbrake.notify_or_ignore(exception, request_data(env))
+      end
+    end
+
+    def ignored_user_agent?(env)
```
vmihailenco: I wonder why this logic is not a part of `Airbrake.notify_or_ignore` method? Like `Airbrake.notify_or_ignore` method can check `user_agent` string and ignore provided exception.

lib/airbrake/rails3_tasks.rb
```diff
@@ -59,6 +59,11 @@ def wait_for_threads
     class ActionDispatch::DebugExceptions; def call(env); @app.call(env); end; end
     class ActionDispatch::ShowExceptions; def call(env); @app.call(env); end; end
 
+    # Override BetterErrors middleware if present
```
vmihailenco: How it is possible that there can be several copies of `BetterErrors` middleware in airbrake namespace? Probably I don't understand how Ruby/Rails works.

teknull:
:+1:

lib/airbrake/better_errors.rb
shifi: Dunno. Probably due to some historical reasons that I am not aware of.
I guess it can be changed like that, but not in this PR.

lib/airbrake/rails3_tasks.rb
shifi: When we run `rake airbrake:test` task, we want clean `STDOUT` (only our stuff) and that's the reason we override few middlewares which can add their own stack trace.

vmihailenco: Document this. And it is not clear why we want clean STDOUT.

shifi: Because we want to show only relevant info to users.
We want to show them configuration and if task passed or not.
There is no purpose to see stack trace and other details which are displayed by other middlewares.

vmihailenco:
Add reference to the issue you try to fix.

vmihailenco:
We you decided go this way? I would expect you to insert Airbrake middleware before BetterErrors middleware, so Airbrake middleware can report exception to Airbrake.

shifi:
Yes I thought that too, but I run into few problems while doing that.
On the other hand, better_errors can be loaded into middleware stack on few different places depending on development setup. I can't cover all possible combinations. If Airbrake Rails middleware is loaded before BetterErrors middleware, it will work as expected, otherwise we will report errors from BetterErrors middleware. Simple as that!

vmihailenco:
You can use `config.middleware.insert_before` - http://guides.rubyonrails.org/rails_on_rack.html

shifi:
Yes I know about `insert_before` and I tried that already. It didn't work if you put `airbrake` gem before `better_errors` gem in `Gemfile`. `BetterErrors` middleware is defined, but it's not loaded into stack and I get an error when I try to insert Airbrake before it. It complains about missing middleware. If you place `airbrake` gem bellow `better_errors` that way it would work but I don't want to tell users how to load their gems.
What is wrong with this approach? It's only used in `development` and it doesn't matter how middleware stack is loaded? I don't get it?

lib/airbrake/better_errors.rb
```diff
@@ -0,0 +1,43 @@
+require 'better_errors'
+
+module BetterErrors
+  # Better Errors' error handling middleware with Airbrake support. Including
+  # this in your middleware stack will report uncatched exceptions to Airbrake
+  # and show a Better Errors error page for exceptions raised below this
+  # middleware.
```
vmihailenco: This comment sounds like I should include this middleware if I use better_errors. But what it does is hacking better_errors middleware to play nicely with Airbrake.

vmihailenco:
>What is wrong with this approach? It's only used in development and it doesn't matter how middleware stack is loaded? I don't get it?

How I am supposed to know that you tried `insert_before` and it requires custom gem order? I bet that in a month none will be able to explain what this code does and why it is written this way.

Overall stack of middlewares is a classic/recommended way to fix issues like this, so I think we should use it. Requiring users to load Airbrake in particular order sounds okay to me since that is how this mechanism works.

shifi:
>How I am supposed to know that you tried `insert_before` and it requires custom gem order?

Because I told you that already:
>I would expect you to insert Airbrake middleware before BetterErrors middleware, so Airbrake middleware can report exception to Airbrake and rethrow it so BetterErrors middleware can show error page.

My answer:
>Yes I thought that too, but I run into few problems while doing that...

...
>I bet that in a month none will be able to explain what this code does and why it is written this way.

So what is wrong with this code? You need more comments? What are you suggesting?

>Requiring users to load Airbrake gem in particular order sounds okay to me since that is how this mechanism works.

If you think this is the way, then please explain it to our users and I will remove this code.
[Here](https://help.airbrake.io/discussions/problems/39629-re-cant-get-airbrake-working-in-rails-problems-39603) is a good place to start. Also some of users placed `airbrake` gem above `better_errors` but `better_errors` middleware still showed up pretty low in [middleware stack](https://github.com/airbrake/airbrake/issues/347#issuecomment-108133583).

vmihailenco:
I suggest to:
- use stack of middlewares and insert Airbrake middleware before better_errors middleware if better_errors middleware is loaded.
- add 1 sentence to the readme, .e.g. "airbrake gem should be loaded after all other gems, so it can insert middleware at right position".

That is how Rails works and devs should be ready to pay the price if they are using it.

shifi:
I'm not for this solution, but I've created a [separate PR](https://github.com/airbrake/airbrake/pull/391) so we can discuss on both. Check description.

zefer:
@shifi :+1: for offering 2 different solutions! I thing #391 is the better option, simple without hidden magic.

PR #391, opened 2015-06-12
Insert Airbrake Rails Middleware after Better Errors Middleware

shifi:
Insert Airbrake Rails Middleware after Better Errors Middleware if Better Errors Middleware defined.

This will work when `airbrake` gem is loaded after `better_errors` gem:
https://gist.github.com/shifi/0e4b30554d1bbce7e808

It will not work if `airbrake` gem is loaded before `better_errors` gem:
https://gist.github.com/shifi/ead7c6a5ede5b943abb2

- [ ] If we decide to go with this solution, add note in readme how to order gems/middlewares

**Note** Placing `airbrake` gem after `better_errors` gem doesn't have to give the same middleware stack like mine, it depends on many things
Example: for [this middleware](https://github.com/airbrake/airbrake/issues/347#issuecomment-108133583), `airbrake` gem is loaded before `better_errors` gem and it has different order than mine middleware

### DIFF
--- a/lib/airbrake/railtie.rb
+++ b/lib/airbrake/railtie.rb
@@ -12,13 +12,16 @@ module Airbrake
 
     initializer "airbrake.middleware" do |app|
 
-      middleware = if defined?(ActionDispatch::DebugExceptions)
-        # Rails >= 3.2.0
-        "ActionDispatch::DebugExceptions"
-      else
-        # Rails < 3.2.0
-        "ActionDispatch::ShowExceptions"
-      end
+      middleware =
+        if !::Rails.env.production? && defined?(::BetterErrors::Middleware)
+          "BetterErrors::Middleware"
+        elsif defined?(::ActionDispatch::DebugExceptions)
+          # Rails >= 3.2.0
+          "ActionDispatch::DebugExceptions"
+        else
+          # Rails < 3.2.0
+          "ActionDispatch::ShowExceptions"
+        end
 
       app.config.middleware.insert_after middleware,
         "Airbrake::Rails::Middleware"


lib/airbrake/railtie.rb
```diff
@@ -12,13 +12,16 @@ class Railtie < ::Rails::Railtie
...
 
-      middleware = if defined?(ActionDispatch::DebugExceptions)
-        # Rails >= 3.2.0
-        "ActionDispatch::DebugExceptions"
-      else
-        # Rails < 3.2.0
-        "ActionDispatch::ShowExceptions"
-      end
+      middleware =
+        if !::Rails.env.production? && defined?(::BetterErrors::Middleware)
```
zefer: Is there any good reason why we can't remove the `if !::Rails.env.production?` test and only check `if defined?(::BetterErrors::Middleware)`?

zefer:
I prefer this solution to #387. It's nice and simple without any hidden magic!

*If* people have difficulties after this fix, we could maybe add a warning message to the output of `rake airbrake:test` when Better Errors is detected.

lib/airbrake/railtie.rb
vmihailenco: +1

vmihailenco:
I like it. Can you add a test?

teknull:
This is all good work.  Thanks for giving this the attention I think it deserves.  Will definitely help those of us who want to fire AirBrake in deployment along with better errors. :+1:

lib/airbrake/railtie.rb
shifi: The only good reason is to make sure we don't break apps in production. Better errors already check if env is different from production before loading their middleware, but I just wanted additional check in our gem.

shifi:
>*If* people have difficulties after this fix, we could maybe add a warning message to the output of `rake airbrake:test` when Better Errors is detected.

@zefer I think most of users have `airbrake` gem above `better_errors` gem in their `Gemfile` since `better_errors` is in development and test group which is **probably** always at the end of the file.
This will [break airbrake installation](https://gist.github.com/shifi/1fa075aaa8d2ece26821) so we can't add warning message because airbrake will never be initialized. Also for apps which already have airbrake, we can't add warning message in rake task since middleware initialization will fail.

shifi:
@zefer ping

Issue #284, opened 2014-04-29
re-enable sending/storing cookies with airbrakes


alexch:
https://github.com/airbrake/airbrake/issues/185 and https://github.com/airbrake/airbrake/commit/c9098b13342de17b252b373c7af94e9f2f30b383 removed cookies, sessions, and some other "sensitive" information from airbrakes. While in some cases this information is important to remove, in others it is important to include. (For instance I just had an issue where a number of customers had an "invalid UTF-8 sequence" in a cookie that my app set; seeing the actual value of that cookie would have helped me track down the bug.)

I suggest enabling sending cookies again by default, and/or adding an explicit configuration setting to enable/disable cookies, rather than the current code using the SENSITIVE_RACK_VARS constant with no (apparent) supported way to modify it.

shifi:
@alexch I will make rack vars filters configurable this week.
I will keep you updated.

kyrylo:
We plan a major release, where by default we send this information. You indeed will be able to filter this information by yourself.

PR #392, opened 2015-06-17
Make rack vars filters configurable

shifi:
Make rack vars filters configurable so users can replace or append to [default list](https://github.com/airbrake/airbrake/blob/master/lib/airbrake/utils/rack_filters.rb).

GH issue: #284 

New docs are ready once we push new gem release.

### DIFF
--- a/lib/airbrake/configuration.rb
+++ b/lib/airbrake/configuration.rb
@@ -6,9 +6,10 @@ module Airbrake
         :development_lookup, :environment_name, :host,
         :http_open_timeout, :http_read_timeout, :ignore, :ignore_by_filters,
         :ignore_user_agent, :notifier_name, :notifier_url, :notifier_version,
-        :params_filters, :params_whitelist_filters, :project_root, :port, :protocol, :proxy_host,
-        :proxy_pass, :proxy_port, :proxy_user, :secure, :use_system_ssl_cert_chain,
-        :framework, :user_information, :rescue_rake_exceptions, :rake_environment_filters,
+        :params_filters, :params_whitelist_filters, :rack_vars_filters,
+        :project_root, :port, :protocol, :proxy_host, :proxy_pass, :proxy_port,
+        :proxy_user, :secure, :use_system_ssl_cert_chain, :framework,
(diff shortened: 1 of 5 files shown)


vmihailenco:
What is the purpose of having separate sets of `:params_filters` and `:rack_filters`? Why `:params_filters` can't contain `:rack_filters`?

>New docs are ready once we push new gem release.

I would like to see docs - they could help me to understand this PR better.

shifi:
It was like that before, I'm not sure why. We also have `rake_environment_filters` which are configurable but don't have default list, which I'm planning to fix to include default list of rack vars.
If we add default rack vars filters to default list of params filters, we could break current behaviour for users who use `config.params_filters = []` or similar.
That way, we wouldn't remove [any rack var from default list](https://github.com/airbrake/airbrake/blob/master/lib/airbrake/utils/rack_filters.rb), and everything would be sent to Airbrake. I think this is the most sensitive data which should be removed always.

>I would like to see docs - they could help me to understand this PR better.

[Docs](https://gist.github.com/shifi/049e3185771d02ac2d0f)

vmihailenco:
I think that is fine. `config.params_filters = []` means that user does not want to use filters. So slightly changing behaviour regarding `rack_filters` seems fine to me.

Please merge :rack_filters into :params_filters and update changelog (do we have it?).

I also think that we should not touch `rake_environment_filters`, deprecate it and remove from the docs if it is documented.

shifi:
@zefer thoughts?

zefer:
I think we should put a hold on this, and rethink filtering as part of a new major release (`v5.0.0`) that will require a new configuration file format. It's time to tidy up this configuration to make it easier to understand. Something along these lines might work...

Use a *single* config variable to define *all* filters (for env vars, params, etc). Use a second config variable that specifies if this is a whitelist or a blacklist. It might look like below...

```rb
# Config defaults.
var_filters = %w[password password_confirmation]
var_filters_mode = :blacklist
```

shifi:
:+1:

shifi:
Although, there would be much more defaults. Current params filters default + rack vars filters default

PR #403, opened 2015-08-14
Fix dumping notice object to xml for rails 2.3.10 and builder 2.1.2

lisovskyvlad:
Hallo. I have rails app (2.3.10), errbit (latest version for today).

In the case if I want generate airbrake notification with Airbrake.notify(e) it sends to errbit wrong xml, without data about error. This patch fix the problem.

### DIFF
--- a/lib/airbrake/notice.rb
+++ b/lib/airbrake/notice.rb
@@ -154,7 +154,7 @@ module Airbrake
           notifier.version(notifier_version)
           notifier.url(notifier_url)
         end
-        notice.error do |error|
+        notice.tag!('error') do |error|
           error.tag!('class', error_class)
           error.message(error_message)
           error.backtrace do |backtrace|


kyrylo:
Hi, @lisovskyvlad. Thanks for your work. 

Does the fix work for you, because `notice.error` is the Builder 3 API and you have Builder 2, which only understands `notice.tag!`?

lisovskyvlad:
@kyrylo No, it is because Builder API 2 has 'error' method, which return empty string (for me) and not xml tag 'error' as expected.

kyrylo:
Hm, that's strange. Does it return `notifier` for you now? https://github.com/lisovskyvlad/airbrake/blob/master/lib/airbrake/notice.rb#L152

lisovskyvlad:
@kyrylo Yeah, It does return all except error block.

kyrylo:
We have released a new version.

lisovskyvlad:
Big thanks!

PR #404, opened 2015-08-15
Fix missing api_key when using V3 endpoint URL

### LINKED ISSUE #364: Bug in building V3 endpoint url
In Sender class you build url for v3 endpoint using format /api/v3/projects/:project_id/notices?key=:api_key, but then in this line https://github.com/airbrake/airbrake/blob/master/lib/airbrake/sender.rb#L54 you fetch only path part on this url, removing ?key=:api_key part.

Api key is not send to the server.

zeroed:
This PR would fix #364 

In `Airbrake::Sender`, using the `URI.path` method [here](https://github.com/airbrake/airbrake/blob/master/lib/airbrake/sender.rb#L54) the resulting URL is not compliant the format `/api/v3/projects/:project_id/notices?key=:api_key`.

IMHO, using [`request_url`](http://ruby-doc.org/stdlib-2.2.2/libdoc/uri/rdoc/URI/HTTP.html#method-i-request_uri) should fix this problem.

#### Issue Type:
- Bug Report

#### Airbrake Gem Version:
4.3.0

#### Ruby Version:

ruby 1.9.3/2.2.1

#### Airbrake Configuration:

```ruby
  config.api_key = "aaaaaaaa"
  config.project_id = "000000"
  config.host = "collect.airbrake.io"
end
```

#### Steps To Reproduce:

From a IRB/PRY session:

```ruby
require 'airbrake' ; 
Airbrake.configure do |c| 
  c.api_key = "000" ;  
  c.environment_name = 'test' ;  
  c.development_environments = []
end

Airbrake.notify(RuntimeError.new("[IGNORE_ME] ..." ))
```

#### Expected Results:

A notification reference

```
# => "1234567890"
```

#### Actual Results:

```
# => nil
```

### DIFF
--- a/lib/airbrake/sender.rb
+++ b/lib/airbrake/sender.rb
@@ -51,7 +51,7 @@ module Airbrake
       http = setup_http_connection
 
       response = begin
-                   http.post(url.respond_to?(:path) ? url.path : url,
+                   http.post(request_uri,
                              data,
                              headers)
                  rescue *HTTP_ERRORS => e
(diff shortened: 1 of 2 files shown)


kyrylo:
Hi, @zeroed. Good stuff!
I'm happy to merge this, but could you please squash your commits into 1?

zeroed:
I'm so glad to help @kyrylo .

Thank you so much for your work...

Issue #410, opened 2015-09-01
rake airbrake:test fail on Rails 5 API only apps


TangRufus:
On a Rails 5 api-only app, `rake airbrake:test` fails
Any work around at the moment?

Info:

```bash
$ foreman run rake airbrake:test
Configuration:
                  api_key: "***********************"
        backtrace_filters: [#<Proc:0x007f98421a3178@/Users/numberthree/.rvm/gems/r
 development_environments: []
       development_lookup: true
         environment_name: "development"
                     host: "api.airbrake.io"
        http_open_timeout: 2
        http_read_timeout: 5
                   ignore: ["ActiveRecord::RecordNotFound", "ActionController::Rou
        ignore_by_filters: []
        ignore_user_agent: []
            notifier_name: "Airbrake Notifier"
             notifier_url: "https://github.com/airbrake/airbrake"
         notifier_version: "4.3.1"
           params_filters: ["password", "password_confirmation", /\w*_token/]
 params_whitelist_filters: []
             project_root: #<Pathname:/Users/numberthree/Documents/ruby-on-rails/p
                     port: 80
                 protocol: "http"
               proxy_host: nil
               proxy_pass: nil
               proxy_port: nil
               proxy_user: nil
                   secure: false
use_system_ssl_cert_chain: false
                framework: "Rails: 5.0.0.alpha"
         user_information: "Airbrake Error {{error_id}}"
   rescue_rake_exceptions: nil
 rake_environment_filters: []
                test_mode: nil
Setting up the Controller.
DEPRECATION WARNING: prepend_before_filter is deprecated and will be removed in Rails 5.1. Use prepend_before_action instead. (called from <class:ApplicationController> at /Users/numberthree/.rvm/gems/ruby-2.2.3@hazy-api/gems/airbrake-4.3.1/lib/airbrake/rails3_tasks.rb:89)
Processing request.


Started GET "/verify" for 127.0.0.1 at 2015-09-02 02:14:34 +0800
Raising 'AirbrakeTestingException' to simulate application failure.
undefined method `to_hash' for {"controller"=>"application", "action"=>"verify"}:ActionController::Parameters
``` 

Gemfile.lock
```
GIT
  remote: git://github.com/rack/rack.git
  revision: 6c4160b8c5173299f4b49ea2c9e4aab76f6b9054
  specs:
    rack (2.0.0.alpha)
      json

GIT
  remote: git://github.com/rails/arel.git
[...]

PR #411, opened 2015-09-01
Fix airbrake/airbrake#410

TangRufus:
For Rails 5 api-only apps, `ActionController::Parameters` has no `to_hash` method.

### DIFF
--- a/lib/airbrake/rails/controller_methods.rb
+++ b/lib/airbrake/rails/controller_methods.rb
@@ -4,7 +4,7 @@ module Airbrake
 
       def airbrake_request_data
         {
-          :parameters       => airbrake_filter_if_filtering(params.to_unsafe_h),
+          :parameters       => airbrake_filter_if_filtering(to_hash(params)),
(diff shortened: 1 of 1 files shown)


kyrylo:
Hi. I'm curious, why `#to_unsafe_h` and not `#to_h`?

TangRufus:
I think `airbrake_filter_if_filtering` is responsible for the filtering. 
And, non-permitted parameters might cause errors, which should be sent to Airbrake.

kyrylo:
I just tested this patch with Rails 3.2.11 and, unfortunately, it doesn't have `#to_unsafe_h`. I think we need to find a better solution.

kyrylo:
Does this work for you?

```
def to_hash(params)
  # Rails <= 4
  return params.to_hash if params.respond_to?(:to_hash)

  # Rails >= 5
  params.to_unsafe_h
end
```

TangRufus:
Added @kyrylo suggestion.

kyrylo:
Thanks!

PR #412, opened 2015-09-04
Remove "API_KEY" set from ENV Variable in Capistrano and airbrake:deploy

midN:
Removed "API_KEY" param being set in Capistrano tasks and airbrake:deploy task
due to possible conflict with ENV and since it's not needed there and is not
mentioned in Docs regarding deployment.

**Bug / Conflict with ENV variable description:**
If you have ENV variable called API_KEY either locally during deployment or
in your application used somewhere else, Airbrake.configuration.api_key will
be ignored and ENV variable called API_KEY used instead.

### DIFF
--- a/lib/airbrake/capistrano.rb
+++ b/lib/airbrake/capistrano.rb
@@ -82,7 +82,6 @@ module Airbrake
             notify_command << " RAILS_ENV=#{rails_env}" if rails_env
             notify_command << " airbrake:deploy TO=#{airbrake_env} REVISION=#{current_revision} REPO=#{repository} USER=#{Airbrake::Capistrano::shellescape(local_user)}"
             notify_command << " DRY_RUN=true" if dry_run
-            notify_command << " API_KEY=#{ENV['API_KEY']}" if ENV['API_KEY']
             logger.info "Notifying Airbrake of Deploy (#{notify_command})"
             if configuration.dry_run
               logger.info "DRY RUN: Notification not actually run."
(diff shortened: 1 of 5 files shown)


mmcdaris:
Hi @midN Thanks for your contribution to the Airbrake gem!
I agree that we should remove `API_KEY` since the `Airbrake.configuration.api_key` is already going to give the best value.

I don't see a good reason to prefer `API_KEY` which may have nothing to do with airbrake and is not documented in wiki doc for [tracking deploys](https://github.com/airbrake/airbrake/wiki/Tracking-deployments-with-Airbrake)

overwriting `Airbrake.configuration.api_key` with whatever value is in env var `API_KEY` would just cause unexpected deploy notification failures that would not be very much fun to troubleshoot.

@shifi do you agree this is safe to remove?

shifi:
@midN @mmcdaris this is a good improvement.
Can you also update [`airbrake_tasks.rb`](https://github.com/airbrake/airbrake/blob/master/lib/airbrake_tasks.rb#L17)?
I don't think we will need this part anymore: `opts.delete(:api_key)`
Thank you.

midN:
@mmcdaris @shifi Thank you for the feedback!
Removed api_key from opts in AirbrakeTasks.deploy and also removed 1 test that was checking if passing api_key opt to AirbrakeTasks.deploy will be used for sending notification.

shifi:
Awesome! Thanks a lot @midN

PR #418, opened 2015-09-14
railtie: insert our Rails middleware before Rails' one

### LINKED ISSUE #401: Track 404 with airbrake
Hi,

How is it possible if airbrake put middleware after `ActionDispatch::ShowExceptions` ?

``` ruby
use Airbrake::UserInformer
use Rack::Sendfile
use #<ActiveSupport::Cache::Strategy::LocalCache::Middleware:0x007f6e413c0058>
use Rack::Runtime
use Rack::MethodOverride
use ActionDispatch::RequestId
use Rails::Rack::Logger
use ActionDispatch::ShowExceptions
use ActionDispatch::DebugExceptions
use Airbrake::Rails::Middleware
```

I think it should be other way around

```
use Rails::Rack::Logger
use Airbrake::Rails::Middleware
use ActionDispatch::ShowExceptions
use ActionDispatch::DebugExceptions
```

kyrylo:
Fixes #401 (Track 404 with airbrake)

> The ActionDispatch::ShowExceptions middleware rescues any exception
> returned by the application and renders nice exception pages if the
> request is local or if config.consider_all_requests_local is set to
> true. If config.action_dispatch.show_exceptions is set to false,
> exceptions will be raised regardless.

Changing the order of middleware allows us to rescue the exception
coming from the Rails middleware. Please note that [we still ignore some
exceptions][1] such as ActionController::RoutingError, which we are able
to catch now.

[1]: https://github.com/airbrake/airbrake/blob/02d29ccde1c46f957c9cf8261835ec7f9386e041/lib/airbrake/configuration.rb#L146

### DIFF
--- a/features/step_definitions/rails_application_steps.rb
+++ b/features/step_definitions/rails_application_steps.rb
@@ -263,5 +263,5 @@ Then /^the Airbrake middleware should be placed correctly$/ do
   airbrake_index  = middleware.rindex("use Airbrake::Rails::Middleware")
   middleware_index = middleware.rindex("use ActionDispatch::DebugExceptions") ||
     middleware.rindex("use ActionDispatch::ShowExceptions")
-  (airbrake_index > middleware_index).should be_true
+  (airbrake_index < middleware_index).should be_true
 end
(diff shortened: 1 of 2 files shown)


kyrylo:
This is broken. Needs more investigation.

Issue #419, opened 2015-09-26
Avoid calling Ruby URI lib


alexis779:
Airbrake middleware makes our Rails app crash on a request parameter containing a |
![uri_parsing_error](https://cloud.githubusercontent.com/assets/210994/10115283/49b70bec-63b7-11e5-8884-5817c11f871b.png)

Sorry for the backtrace as an image.

We can not control the parameter sent by HTTP client so we need to handle this scenario.

URI.join raise an exeception when request parameter has unescaped special characters like pipe:

```
[1] pry(main)> URI.join("http://www.example.com", "/|")
URI::InvalidURIError: bad URI(is not URI?): /|
from /usr/lib/ruby/2.1.0/uri/generic.rb:1203:in `rescue in merge'
```

but we don't need to call it to build the full url of the request.

alexis779:
See https://github.com/airbrake/airbrake/pull/420

kyrylo:
Fixed by https://github.com/airbrake/airbrake/pull/420.

PR #420, opened 2015-09-26
[#419] don't fail on request param containing |

alexis779:
Replace URI.join

### DIFF
--- a/lib/airbrake/rails/controller_methods.rb
+++ b/lib/airbrake/rails/controller_methods.rb
@@ -98,7 +98,7 @@ module Airbrake
           url << ":#{request.port}"
         end
 
-        URI.join(url, request.fullpath).to_s
+        url << "/" << request.fullpath
       end
 
       def airbrake_current_user
--- a/test/controller_methods_test.rb
+++ b/test/controller_methods_test.rb
@@ -36,6 +36,14 @@ class NoSessionTestController < TestController
   end
 end
 
+class ParamTestController < TestController
+  QUERY_PARAMS = { name: "|" } # URI parameter with invalid character
+  def request
+    query = QUERY_PARAMS.map { |k, v| "#{k}=#{v}"}.join("&")
+    OpenStruct.new(:port=> 80, :protocol => 'http://', host: 'example.com', :fullpath => "path?#{query}", :env => [])
+  end
+end
+
 class ControllerMethodsTest < Test::Unit::TestCase
   include DefinesConstants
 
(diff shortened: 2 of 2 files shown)


kyrylo:
Hi, @alexis779!

Thanks for this PR, but I think it's better to `escape` fullpath like this:

```ruby
URI.join(url, URI.escape(request.fullpath)).to_s
```

Could you please confirm that this works for you?

I'd also be happy to see a test that tests `|`.

kyrylo:
@alexis779 a gentle ping.

alexis779:
The problem with URI.escape is that it modifies the input sent by the client. For example `|` is rewritten into `%7C`.
Feel free to adapt the PR if you have a good reason to keep that.

I added a unit test that, without the PR, fails with

```
Error: test: #airbrake_request_url should handle invalid character in param. (ControllerMethodsTest): URI::InvalidURIError: bad URI(is not URI?): path?name=|
/usr/lib/ruby/2.1.0/uri/generic.rb:1203:in `rescue in merge'
/usr/lib/ruby/2.1.0/uri/generic.rb:1200:in `merge'
/usr/lib/ruby/2.1.0/uri/common.rb:237:in `each'
/usr/lib/ruby/2.1.0/uri/common.rb:237:in `inject'
/usr/lib/ruby/2.1.0/uri/common.rb:237:in `join'
/usr/lib/ruby/2.1.0/uri/common.rb:785:in `join'
/home/alexis/job/airbrake/lib/airbrake/rails/controller_methods.rb:102:in `airbrake_request_url'
/home/alexis/job/airbrake/test/controller_methods_test.rb:135:in `block (2 levels) in <class:ControllerMethodsTest>'
/var/lib/gems/2.1.0/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `instance_exec'
/var/lib/gems/2.1.0/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `block in create_test_from_should_hash'
```

Better than a Newrelic screenshot ...


Also you can manually reproduce the error in a new Rails app adding `thin` (`webrick` would send a BadRequest) and `airbrake` in the Gemfile
```
gem 'thin'
gem 'airbrake'
```
with airbrake initializer `config/initializers/airbrake.rb`
```
  config.api_key = '<KEY>'
end
```

kyrylo:
I can see your point about the modified input. Thanks for the PR!

kyrylo:
Sorry, I had to revert this change, because it fails our build. At the moment I cannot share the link to the failed build, but I'm resolving this now.

kyrylo:
https://circleci.com/gh/airbrake/airbrake/314

Issue #413, opened 2015-09-08
Rails 5 deprecation warning


TangRufus:
Rails 5 needs actual class references instead of strings or symbols.

```
01:37:52 log.1  | DEPRECATION WARNING: Passing strings or symbols to the middleware builder is deprecated, please change
01:37:52 log.1  | them to actual class references.  For example:
01:37:52 log.1  |
01:37:52 log.1  |   "Airbrake::Rails::Middleware" => Airbrake::Rails::Middleware
01:37:52 log.1  |
01:37:52 log.1  | . (called from block in tsort_each at /Users/******/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/tsort.rb:226)
01:37:52 log.1  | DEPRECATION WARNING: Passing strings or symbols to the middleware builder is deprecated, please change
01:37:52 log.1  | them to actual class references.  For example:
01:37:52 log.1  |
01:37:52 log.1  |   "Airbrake::UserInformer" => Airbrake::UserInformer
01:37:52 log.1  |
01:37:52 log.1  | . (called from block in tsort_each at /Users/*******/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/tsort.rb:226)
```

kyrylo:
Hey, thanks for testing Airbrake with Rails 5. I appreciate that a lot. I'd be happy to merge a fix, if you're interested :)

TangRufus:
pull request #422 created

kyrylo:
Closing. See the PR for the reason.

PR #422, opened 2015-10-04
Avoid Rails 5 deprecation warning

TangRufus:
Passing strings or symbols to the middleware builder is deprecated since Rails 5

Fix airbrake/airbrake#413


kyrylo:
Hey, thanks for the PR, but from my understanding this will break support for older Rails versions. We do want to keep supporting for them.

kyrylo:
Hey, I think we're going to close this. We're preparing a new major release of this library, which will include Rails 5 support.

PR #429, opened 2015-10-08
Corrected ControllerMethods test that could never fail

iainbeeston:
This test was previously checking the wrong thing. It was calling a
private method to load the user (`#airbrake_current_user`), then checking
that `#id` is never called on that user. However,
`#airbrake_current_user` never calls `#id` even when a valid user is
loaded. Therefore this test could never fail.

I've "corrected" it by making it call `#airbrake_request_data` instead,
which does attempt to call `#id` (this is also the method under test for
every other test in this context).

### DIFF
--- a/test/controller_methods_test.rb
+++ b/test/controller_methods_test.rb
@@ -39,7 +39,7 @@ end
 class ControllerMethodsTest < Test::Unit::TestCase
   include DefinesConstants
 
-  context "#airbrake_current_user" do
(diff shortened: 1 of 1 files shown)


kyrylo:
This test started to make no sense after [that](https://github.com/airbrake/airbrake/commit/245f5b949a95c6bf916901d4c4c9c5e17fd62161) commit. Good spot!

PR #430, opened 2015-10-08
Protected against errors while gathering user details

iainbeeston:
When airbrake reports a controller error, it attempts to get information
about `current_user`. Airbrake assumes that getting `current_user` is a
safe operation. If there is an error while loading `current_user`
airbrake does not handle the error, and crashes. This both masks the
original error with a user loading error, and prevents either error
from being reported to airbrake.

This change rescues errors that occur when calling `current_user` or
`current_member`. If this happens, the user info sent to airbrake will be
empty.

We also handle any errors raised while attempting to fetch user defined
attributes on `current_user`. In this case the erroneous attribute is
not included in the user info, but any attributes that are successfully
fetched will still be included.

### DIFF
--- a/lib/airbrake/rails/controller_methods.rb
+++ b/lib/airbrake/rails/controller_methods.rb
@@ -10,12 +10,12 @@ module Airbrake
           :action           => params[:action],
           :url              => airbrake_request_url,
           :cgi_data         => airbrake_filter_if_filtering(request.env),
-          :user             => airbrake_current_user || {}
+          :user             => airbrake_current_user
         }
       end
 
       private
-      
+
       def to_hash(params)
         # Rails <= 4
         return params.to_hash if params.respond_to?(:to_hash)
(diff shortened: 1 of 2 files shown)


kyrylo:
Hi, @iainbeeston! 
I'm still not quite sure why we need to rescue here. What are realistic scenarios when `current_user` fails? What do you mean by "broken implementation"?

iainbeeston:
This is happening in a real world app that I maintain. In a rails controller, we had an error that was raised in a before filter\*, which was run *before* the code to configure user authentication\*\*. When airbrake then called `airbrake_request_data`, the call to `current_user` raised an error, and airbrake itself failed.

But surely, as `current_user` is an unknown method (as far as airbrake is concerned) it could fail for any number of reasons (what if it loads the current user from the database and raises an error if it cannot find the user?). The same goes for getting attributes from the user model to put into the request data.


\* Validating csrf tokens on form submissions
\*\* We use authlogic. Authlogic has a before filter to configure auth methods on your controller before you can get the current user from the session. If you try to load the user before this point the controller does not know how to handle auth.

kyrylo:
Thanks, I see your reasoning.

kyrylo:
We just released a new version (4.3.2) with your patch included. Thanks!

iainbeeston:
Great - I'll try it out soon.

Also credits should go to @JoelQ as well who paired with me on this :-)

PR #432, opened 2015-10-19
[Fix] filtering multidim-arrays from cgi-data's 'action_dispatch.cookies'

loybert:
## Details
Issue Type: Bug Report
Airbrake Gem Version: current (4.3.2)
Ruby Version: 2.3.2
Framework Name/Version: Rails 4.2.4
Airbrake Configuration: basic (only api-key in initializer)

## Steps To Reproduce:
Having cookies set with key, that should be filtered.

## Expected Results:
It should filter the cookie value. But in any case airbrake should not break (so it notifies at least about ANY error).

## Actual Results:
Airbrake breaks on params-filtering and nobody notices.

# PullRequest fix details
When cleaning cgi-data in `utils/params_cleaner.rb`, there might be cookies set on the key `action_dispatch.cookies` (Rails 4.2.4), which contain a filter-key. 

E.g.: 
```
# key to be filtered
@blacklist_filters = %(secret)

# cgi hash, including cookies
cgi_data = {
  'action_dispatch.cookies' => [
    ['_session', 'htRkVDRGI1NEpNcElVNGZSdHRUR0hpdUxlWWJnZjRqUjFVaUVwNU1nSW9'],
    ['secret', 'usually this value would be filtered'],
  ]
}
```

This cgi-data would break the params-cleaner, when it's attempting to write a hash-key onto an array (see `utils/params_cleaner.rb` line 102). Even worse, this breakage hides any raised exception from the application and usually nobody would notice if no other error-logger is active.

I also assume, avoiding this error-case was also intended by #308. The only difference in my PR is skipping a non-hash on the `if` statement, but allowing deeper lookup in the last `elsif` statement.

### DIFF
--- a/lib/airbrake/utils/params_cleaner.rb
+++ b/lib/airbrake/utils/params_cleaner.rb
@@ -98,7 +98,7 @@ module Airbrake
 
         def filter(hash)
           hash.each do |key, value|
-            if filter_key?(key)
+            if hash.is_a?(Hash) && filter_key?(key)
               hash[key] = "[FILTERED]"
             elsif value.respond_to?(:to_hash)
               filter(hash[key])
--- a/test/params_cleaner_test.rb
+++ b/test/params_cleaner_test.rb
@@ -190,4 +190,15 @@ class ParamsCleanerTest < Test::Unit::TestCase
     assert_match(/\A#<(Temp)?[Ff]ile:0x.+>\z/, clean_params.parameters[:files][0])
     assert_match(/\A#<IO:0x.+>\z/, clean_params.parameters[:files][1])
   end
+
+  should "not break on filtering multi-dimensional array as possible in action_dispatch.cookies" do
+    original = { 'cgi_cookies_to_filter' => [['any_cookie_key', 'some_cookie_value'], ['secret', 'some_secret_value']] }
+    clean_params = clean(:params_filters => [:secret],
+                         :params_whitelist_filters => [:secret],
+                         :parameters => original)
+    assert_nothing_raised do
+      clean_params.send(:parameters)
+    end
+  end
+
 end


kyrylo:
Hello. Thanks for the PR!

I'm not sure I understand your fix. I can see you left a comment about the "desired behaviour". I agree that the behaviour your PR introduces is confusing. It feels to me like you're amending the actual data instead of filtering it.

Could you please explain why `[['filter_me', 'some_value'], ['do_not_filter_me', 'some_other_value']]` becomes `[['filter_me', '[FILTERED]'], ['do_not_filter_me', 'some_other_value']]`?

To me it feels like `filter_me` is not a key, and `some_value` is not a value associated with `filter_me`, so if you aim to filter anything here, it should be `'cgi_cookies_to_filter'`.

loybert:
hi @kyrylo , sorry in advance for my confusing spec params. 
Actually there are two issues involved on that spec. First of all:

a) `clean_params` should not break the entire Airbrake notification, when a multi-dimensional array gets processed, which contains any filter-key as value.
In this spec the decisive point is the variable `original` containing the String `filter_me` as value inside of any array `original = { 'cgi_cookies_to_filter' => [['filter_me',  ...`
So an array like this shouldn't lead to an error. That's what my spec and my fix are for.

The other issue, is how to handle sensitive data inside of cgi cookie-data, as in the example above from `action_dispatch.cookies` (see headline *PullRequest fix details* in first post). 
b) Being naive as a user: I would expect airbrake filters the sensitive values of certain cookies, depending how I configured my filter-keys. But especially with those cookies, there isn't any Hash-object which contains my key-value pairs, but it's an array with multiple sub-arrays containing two values (in manner of a key value pair). Just because those `action_dispatch.cookies` are set that way, doesn't mean either **every** Array with exactly two values, represents a key-value pair. So doing any filtering there (what i alluded in my `filter` variable inside of the spec) would be just a wild guess and could mess up notification-payloads because of false interpretation of a value being a key.
So handling `action_dispatch.cookie` wouldn't work generically and needs some kind of pre-processing when processing `clean-cgi-data`. This is not handled by this PR, it is just kindly mentioned. Because wether filtering cookies like those, should be handled or not, is still your decision. :-)

Kind regards Daniel

loybert:
```
def filter(hash)
  hash.each do |key, value|
    if filter_key?(key)
      hash[key] = "[FILTERED]"
    elsif value.respond_to?(:to_hash)
      filter(hash[key])
    elsif value.is_a?(Array)
      hash[key] = value.inject(Array.new) do |result, item|
      item = filter(item) if item.is_a?(Enumerable)
      result.push(item)
    end
  end
end
```

When passing `hash = [['foo', 'bar'], ['password', '123']]` to this method, the last `elsif` statement would match and calls `filter` on the sub-array `['foo', 'bar']` as hash recursively. Running `['foo', 'bar'].each |key, value|` results in two iterations with:
iteration 1 => `key = 'foo'`, `value = nil` 
iteration 2 => `key = 'bar'`, `value = nil` 

But wether `'foo'` neither `'bar'` are filter-keys, so no case of that if-elseif block matches.


Processing the recusion with the second sub-array `['password', '123']`, the iterations would be:
iteration 1 => `key = 'password'`, `value = nil` 
iteration 2 => `key = '123'`, `value = nil` 

On the first iteration the if-statement `if filter_key?(key)` matches (because `'password'` might be something sensitive) and in the next line it tries to write a key on an array named `hash`.

```
# iteration 2, right after => if filter_key?(key)
hash = ['password', '123']
key = 'password'

hash[key] = '[Filtered]' # => TypeError!
```


Problem of this recursion (`item = filter(item) if item.is_a?(Enumerable)` ) is the statement `if item.is_a?(Enumerable)` is not sufficient enough, to assure `item` is a hash (an array also matches this condition). But for this moment this is fine, because we just want to go deeper into this object, to find even more filterable values. But when entering the `filter` method on the first recursive step, we handle the param `hash` blindly as a Hash, before we made sure it is really one (and not an array).
Running `each do |key, value|` also works on an array, so the confusion is perfect.

Hope my debugging example helps :-)
    if filter_key?(key)

kyrylo:
> `clean_params` should not break the entire Airbrake notification, when a multi-dimensional array gets processed, which contains any filter-key as value.

I totally agree with this.

> PS: the `filtered` variable inside the spec isn't used. The spec only expects not running into an error.

Thanks, I hadn't notice that when I wrote my comment.

Thanks for the extensive description. So just to be 100% sure, I'd like to clarify: does your PR actually modify `original`? If that's true, then I don't think it should work like so. If you're interested in cookies not being sent to Airbrake, you could use a [custom filter](https://github.com/airbrake/airbrake/wiki/Filtering-or-ignoring-data-before-it's-sent-to-Airbrake#ignore-by-custom-conditions).

If it doesn't modify it, I am happy to merge this, but please delete unused code and squash your commits into 1 :)

loybert:
> Thanks for the extensive description. So just to be 100% sure, I'd like to clarify: does your PR actually modify original? If that's true, then I don't think it should work like so. If you're interested in cookies not being sent to Airbrake, you could use a custom filter.

No, it doesn't modify original. It is just continuing the recursions, in manner to find another hashes again in deeper levels (which are able to be filtered).

> If it doesn't modify it, I am happy to merge this, but please delete unused code and squash your commits into 1 :)

Done ;-)

kyrylo:
Thanks!

loybert:
thanks for merging :bow:

kyrylo:
By the way, we plan to release a new version of the library today. I will let you know.

kyrylo:
Airbrake 4.3.3 is out!

Issue #370, opened 2015-03-07
Airbrake not respecting config.ignore on wrapped exceptions


joevandyk:
Rails raises `ActionDispatch::ParamsParser::ParseError`that wraps a `JSON::ParserError` exception. I have `config.ignore << "ActionDispatch::ParamsParser::ParseError"` in my Airbrake config.

However, the exceptions are still being sent to Airbrake, it shows up as `JSON::ParserError`.

I don't want to ignore all `JSON::ParserErrors`, just this particular one while it's wrapped in the ActionDispatch one.

kenn:
Came from https://github.com/rails/rails/issues/17731 and true, it's an Airbrake bug.

Also, I think it's worth considering that client errors like `ActionDispatch::ParamsParser::ParseError` or `ActionController::ParameterMissing` can be ignored by default.

joevandyk:
Yes, I think those should be added to the default list. 

I have a pull request out there for fixing this bug btw.

kenn:
Any thoughts? @dvdplm @shifi @alif

kenn:
FWIW, it's a win-win because the Airbrake server will have fewer API hits, and users see less noise and have less time wasted.

shifi:
Sorry guys for waiting so long. I will check this soon.

PR #433, opened 2015-10-20
Add way to ignore all the exceptions

joevandyk:
fixes https://github.com/airbrake/airbrake/issues/370


kyrylo:
We just released a new version with your patch. Thanks! :)

PR #434, opened 2015-11-16
Refactor

markburns:
This tackles some of the largest issues highlighted in code climate.

https://codeclimate.com/github/airbrake/airbrake/Airbrake::Notice

Brings this class up from an F to a B.
https://codeclimate.com/github/markburns/airbrake/lib/airbrake/notice.rb?from_sha=68bec497&to_sha=1e706a9f

Overall score goes from 2.83 to 3.66 GPA

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,7 @@
+languages:
+   Ruby: true
+exclude_paths:
+  - "test/*"
+  - lib/airbrake/notice/xml_builder.rb
+  - .codeclimate.yml
+
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -3,5 +3,6 @@ languages:
 exclude_paths:
   - "test/*"
   - lib/airbrake/notice/xml_builder.rb
+  - lib/airbrake/notice/json_builder.rb
   - .codeclimate.yml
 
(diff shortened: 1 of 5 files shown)


kyrylo:
Hello! Thanks for your PR! 

I am looking at this counter `+262 −152` and the changes, and I must say I'm not sure I catch the goal of this refactoring. Yes, the CodeClimate score may be improved, but now it's much harder to support and understand the code after the presented changes. The PR adds 3 new classes, defines inheritance, scatters the code across multiple files, defines helper methods and defines 1 factory method. We also use complex metaprogramming techniques now. This looks a lot more complex than just defining things inline.

I believe it's better to keep it simple, even for the sake of the CodeClimate score.

markburns:
That's one way of thinking of clearer separation of concerns, shorter methods and general reduction in cyclomatic complexity. If you find this code harder to read than the original then I won't try to convince you that it is better.

kyrylo:
@markburns I am very sorry, I appreciate your work. In my opinion the main problem is not solved. The render methods are still huge. That's the real problem.

markburns:
I'd say that the render methods and any other DSL that represents data can be considered a valid place for ignoring cyclomatic complexity, hence the refactor and `.codeclimate.yml` ignore. 
They are effectively views or templates. I don't see how this _doesn't_ solve that issue as that's specifically what I tackled. Unless you don't like to consider them as view-like objects.

I think our style of coding and ways of thinking about what constitutes readable, maintainable code are coming from completely different ways of thinking and I can't empathise with yours, so I can't really contribute anything.

PR #438, opened 2015-12-11
Only notify Airbrake of deployment from primary server

nick-desteffen:
Currently when we do a deployment all servers are triggering the deployment notification to Airbrake.  This really only needs to happen once.


fastman:
Wow, your timing is perfect, I was about to push similar PR, :+1: for this :-)

kyrylo:
Thanks for the PR! What happens if there's no `:primary` role? From my understanding in order to make the presented code work, you have to do this:

```
role :db,  'somedomain.com', :primary => true
```

nick-desteffen:
Good catch.  Since you guys just did a new release I thought it would be cleaner to just open a new pull, so here is the updated pull with your suggestions:
https://github.com/airbrake/airbrake/pull/443

PR #439, opened 2015-12-16
Provide information about where to find the old code

zefer:
This is to minimise confusion for people who are looking for the old
code and its README.


zefer:
Closed in favour of #440 (the original target branch has gone).

PR #443, opened 2015-12-19
Only notify Airbrake of deployment from primary server

nick-desteffen:
Currently when we do a deployment all servers are triggering the deployment notification to Airbrake. This really only needs to happen once.

### DIFF
--- a/lib/airbrake/capistrano/tasks.rb
+++ b/lib/airbrake/capistrano/tasks.rb
@@ -6,7 +6,8 @@ if defined?(Capistrano::VERSION) &&
   namespace :airbrake do
     desc "Notify Airbrake of the deploy"
     task :deploy do
-      on roles(:all) do
+      role = roles(:all, select: :primary).first || roles(:all).first
+      on role do
         within release_path do
           with rails_env: fetch(:rails_env, fetch(:stage)) do
             execute :rake, <<-CMD


lib/airbrake/capistrano/tasks.rb
```diff
@@ -6,7 +6,8 @@
   namespace :airbrake do
     desc "Notify Airbrake of the deploy"
     task :deploy do
-      on roles(:all) do
+      role = roles(:all, select: :primary).first || roles(:all).first
```
kyrylo: I'm not sure why you need `.first` (it wasn't needed before). Could you please explain?

nick-desteffen: `.first` is needed in the event that `roles(:all, select: :primary)` returns an empty array, (no primary server was defined, which was your concern with the previous pull request).  The first element will be `nil`, which will then trigger the server to be pulled from the full listing (`rolles(:all)`).  I hope that makes sense.

kyrylo: Thanks for the explanation, that does make sense. That said, how confident are you in this? Why not to do:

```ruby
# If you want to opt in this new behaviour, you define your `:primary`,
# otherwise we just fall back to how it worked before.
roles(:all, select: :primary).first || roles(:all)
```

I'm just surprised that it wasn't an issue for all these years, so `roles(:all).first` _seems_ to be a safer solution.

nick-desteffen:
```ruby
roles(:all).first
``` 
would likely work fine, I think the point is we want it only notifying on 1 server, not all of them.  It was doing it that way at one point, however I'm not sure when it changed, I just noticed when I realized none of our projects were notifying Airbrake anymore, something that must have been missed in our upgrade to Capistrano 3.x.  Choosing the primary server to do the task seems to be the standard, so that's why I went with that.  I can however change it to just be `roles(:all).first` if that's what you would prefer.  Let me know and I'll update the pull request.

kyrylo: Have you tested this? It looks like the `on` method expects an Array (and we now pass a single item)

* https://github.com/capistrano/capistrano/blob/d947b25aa2f788a9cf083ab3d8c2e92675507740/lib/capistrano/dsl.rb#L52-L53
* https://github.com/capistrano/capistrano/blob/e30516859878e510c14f24bfbe14abe94b20bd03/lib/capistrano/configuration.rb#L124

nick-desteffen:
Yes, I tested it with both scenarios and it worked as expected.

On a side note the documentation on https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md#capistrano is incorrect.  You **do** need to define the hook in order to trigger the deployment notification with Capistrano 3

```ruby
namespace :deploy do
  after :finishing, 'airbrake:deploy'
end
```

kyrylo:
Thanks for the PR! I've updated the migration guide.

We just pushed the new gem (v5.0.1).

:+1:

PR #446, opened 2015-12-21
README,Migration_guide: fix Capistrano 3 instructions

kyrylo:
I'm not sure why I considered the hook is not needed. Even if it
worked without the hook for me, turns out defining it is a much
more robust solution.

Thanks to @nick-desteffen for [pointing this out][1]

[1]: https://github.com/airbrake/airbrake/pull/443#issuecomment-166363714

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,6 +342,16 @@ Capfile:
 require 'airbrake/capistrano/tasks'
 ```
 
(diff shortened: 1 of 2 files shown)


mmcdaris:
LGTM!

kyrylo:
:+1:

PR #451, opened 2015-12-28
Update Migration_guide_from_v4_to_v5.md

liorsion:
filter_chain.rb#env_filter receives two parameters: environment, ignore_environments, however - config (which comes from user config) never initializes environment parameter which is always null, and so airbrake receives development errors - this is a crud and rude workaround until you (I suggest) force feed config with the actual env

### DIFF
--- a/docs/Migration_guide_from_v4_to_v5.md
+++ b/docs/Migration_guide_from_v4_to_v5.md
@@ -125,6 +125,7 @@ test_mode | removed | n/a
 
   # New way
   Airbrake.configure do |c|
+    c.environment = Rails.env
     c.ignore_environments = %w(development test)
 
     # OR to collection exceptions in all envs


kyrylo:
Hey! I'm not sure I see what to fix. If you don't configure your environment, then the `ignore_environments` option is useless, because you have no environments.

Our Rails generator generates the line you added in this PR for you.

liorsion:
It doesn't. That's the point. I do configure rails.env otherwise this
wouldn't have worked

kyrylo:
Sorry, I'm still not sure what the problem is. I'm pretty sure our Rails generator [generates this line](https://github.com/airbrake/airbrake/blob/4df0447b18145a6a44d6a56b43e1dae9e233f30a/lib/generators/airbrake_initializer.rb.erb#L42).

Or is it the migration guide section that is confusing?

liorsion:
Ah got it I think. The migration guide didn't say to re run generators if I
was migrating an existing setup.. Or did I miss it?

kyrylo:
No, you are correct, it's not stated anywhere. I think I understand what you struggle with now. This is how it was designed (intentionally).

Thanks for the PR! I'm going to merge it.

PR #460, opened 2016-01-05
circle: update build matrix to test against latest Ruby versions

kyrylo:


vmihailenco:
It is kinda strange that we test only 2.3.0 and 1.9.3...

kyrylo:
Otherwise the build time is pretty damn slow. We don't parallelise enough, though. This is something I am planning to add in the future (we could add more Rubies).

vmihailenco:
LGTM

PR #464, opened 2016-01-05
Explained how to preserve the old value for development environments

iainbeeston:
Between Airbrake v4 and v5 the default environments that are ignored has changed from `[:development, :test, :cucumber]` to `[]`. This means that any apps which did not explicitly define `ignore_environments` before and upgrade to v5 will suddenly start sending errors to airbrake in development and test.


kyrylo:
:+1:

PR #466, opened 2016-01-06
rack: report current_user if it's available

kyrylo:
This change indirectly adds support for [OmniAuth][1] users. The way
OmniAuth integrates with Rails is described [here][2]. Note: this only
works for Rails apps, which I believe should be sufficient for now.

[1]: https://github.com/intridea/omniauth
[2]: https://github.com/testbrian/omniauth-github/blob/patch-2/README.md

### DIFF
--- a/lib/airbrake/rack/notice_builder.rb
+++ b/lib/airbrake/rack/notice_builder.rb
@@ -62,7 +62,7 @@ module Airbrake
           context[:action] = @controller.action_name
         end
 
-        notice[:context].merge!(@user.to_hash) if @user
+        notice[:context].merge!(@user.as_json) if @user
 
         nil
       end
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -1,30 +1,31 @@
 module Airbrake
   module Rack
     ##
-    # Represents an authenticated Warden user, which can be converted to
-    # Airbrake's payload format.
+    # Represents an authenticated user, which can be converted to Airbrake's
+    # payload format. Supports Warden and Omniauth authentication frameworks.
     class User
-      # Finds the Warden user in the Rack environment and creates a new user
-      # wrapper.
+      # Finds the user in the Rack environment and creates a new user wrapper.
       #
       # @param [Hash{String=>Object}] rack_env The Rack environment
       # @return [Airbrake::Rack::User, nil]
       def self.extract(rack_env)
-        return unless (warden = rack_env['warden'])
-        return unless (warden_user = warden.user(run_callbacks: false))
-        new(warden_user)
+        # Warden support (including Devise).
+        if (warden = rack_env['warden'])
+          if (user = warden.user(run_callbacks: false))
+            return new(user) if user
+          end
+        end
+
+        # Fallback mode (OmniAuth support included). Works only for Rails.
+        controller = rack_env['action_controller.instance']
+        new(controller.current_user) if controller.respond_to?(:current_user)
       end
 
-      ##
-      # @param [Warden::Proxy] warden_user
-      def initialize(warden_user)
-        @warden_user = warden_user
+      def initialize(user)
+        @user = user
       end
 
-      ##
-      # Converts the user to Airbrake payload user.
-      # @return [Hash{Symbol=>String}] the hash with retrieved user details
-      def to_hash
+      def as_json
         user = {}
 
         user[:id] = try_to_get(:id)
@@ -39,7 +40,7 @@ module Airbrake
       private
 
       def try_to_get(key)
-        String(@warden_user.__send__(key)) if @warden_user.respond_to?(key)
+        String(@user.__send__(key)) if @user.respond_to?(key)
       end
 
       def full_name
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Added better support for user reporting for Rails applications (including
+  OmniAuth support) ([#466](https://github.com/airbrake/airbrake/pull/466))
+
 ### [v5.0.2][v5.0.2] (January 3, 2015)
 
 * Fixed the bug when Warden user is `nil`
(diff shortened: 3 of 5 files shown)


lib/airbrake/rack/notice_builder.rb
```diff
@@ -62,7 +62,7 @@ def add_context(notice)
           context[:action] = @controller.action_name
         end
 
-        notice[:context].merge!(@user.to_hash) if @user
+        notice[:context].merge!(@user.data) if @user
```
vmihailenco: Why data instead of to_hash?

lib/airbrake/rack/user.rb
```diff
@@ -1,30 +1,33 @@
...
-        new(warden_user)
+        # Warden support (including Devise).
+        if (warden = rack_env['warden'])
+          if (user = warden.user(run_callbacks: false))
+            return new(user) if user
+          end
+        end
+
+        # Fallback mode (OmniAuth support included). Works only for Rails.
+        controller = rack_env['action_controller.instance']
```
vmihailenco: Is `action_controller.instance` always available?

lib/airbrake/rack/user.rb
```diff
@@ -1,30 +1,33 @@
...
+        return unless controller.respond_to?(:current_user)
+        user = controller.current_user
+        new(user)
```
vmihailenco: `new(controller.current_user) if controller.respond_to?(:current_user)` instead of those 3 lines.

svoop:
@kyrylo Thanks for the PR! It doesn't work for my app, the user* info is not included in the payload. I can dig deeper into where it fails and whether it's a "my app only" issue this evening (CET).

kyrylo:
Hey, @svoop! That would be great. I appreciate that!

I've published the repo I was using to test this PR: https://github.com/kyrylo/airbrake-466
Perhaps, your app is configured differently, so that's why it's not working. I hope we can resolve this because without your help I think I won't be able to offer a better solution.

lib/airbrake/rack/notice_builder.rb
kyrylo: Technically, this is a serializer. I checked out how people usually name this method and found out that [Django uses `data`](http://www.django-rest-framework.org/api-guide/serializers/). That was convincing, so I decided to rename.

lib/airbrake/rack/user.rb
kyrylo: Looks like the answer is [yes](https://github.com/rails/rails/blob/762f7daf55a91d2a20bedb4906c9dd04e031d74e/actionpack/lib/action_controller/metal.rb#L203).

kyrylo: And it doesn't matter, because we use `respond_to?`, so `nil.respond_to?(:current_user)` is always `false`.

kyrylo: So it saves us from errors in other frameworks like Sinatra (remember, this file is for Rack frameworks, not just Rails).

lib/airbrake/rack/user.rb
kyrylo: Refactored.

svoop:
@kyrylo My bad, the deploy didn't really go through and in the haste this morning, I failed to see that. Sorry for the noise!

lib/airbrake/rack/notice_builder.rb
vmihailenco: Django conventions in Rails app is interesting :)

vmihailenco:
LGTM

kyrylo:
@svoop thanks for the info. Let me know when you tested this :)

lib/airbrake/rack/notice_builder.rb
zefer: I like your thinking, Kyrylo, but I think the `to_hash` convention would be quicker for the reader to understand, in a Ruby context.

kyrylo: Why not? The concept stays the same for every language.

zefer: Because this is Ruby, and this is not a Ruby convention. I don't insist, it's minor, but I think `data` is less readable because I have to look-up the method to see what it returns, unlike with `to_hash`.

kyrylo: What about `as_json`?

zefer: Curious, why not `to_hash`? I think I prefer that to `as_json` because I always thought `as_json` is a slight misnomer, but it's a convention, and it's fine with me :smile:

kyrylo: I guess `as_json` would be better because with `to_hash` I made [this](https://github.com/airbrake/airbrake/pull/466/files#diff-44014eb454f65244a394186265a92a2cL64) mistake. I was calling `{}.to_hash` with always returns itself.

kyrylo: I agree, it's minor, but `as_json` would be least error prone in my opinion.

zefer: Cool.to_:+1:

kyrylo: Replaced with `as_json`.

svoop:
@kyrylo Your PR works fine here, thanks a bunch!

kyrylo:
@svoop, fantastic! Big thanks for testing it! :+1:

svoop:
Great job, @kyrylo, hope this is worthy of a 5.0.3 release :rocket:

lib/airbrake/rack/user.rb
espen: Shouldn't this be env instead of rack_env if not using Warden? I have a Rails app with current_user in my controller but airbrake is unable to locate it because it tries to fetch rack_env instead of env.

kyrylo: Sorry, I don't think I follow. We're passing `env` from our middleware and call it `rack_env` for clarity. What do you mean by `env`?

Issue #470, opened 2016-01-10
Integration with Phusion Passenger


mvastola:
I'd very much like to see integration between `airbrake` and Phusion's `passenger`, because a major hole in my error notification setup occurs when I deploy to a server and `passenger` fails to spawn instances of my `rails` app (well, technically it treats everything as a `rack` app, but that's beside the point) for whatever reason. 

While `airbrake` already has `rack` integration, this doesn't come into play until/unless `rack` successfully loads. Issues from missing gems to corrupt configurations can prevent `rack` from getting to this point, and currently, with `passenger`, the only way to monitor these errors is to tail your web server log file or subscribe to their (paid) UnionStation service.

In the past half-hour I just scraped together an extremely bare-bones integration as its own gem called [airbrake-passenger](https://github.com/Partyista/airbrake-passenger). Rather than a separate gem, however, I would really like to see this code become part of `airbrake` though.

Would `airbrake` be willing to merge this implementation into their official gem/repository if I refactored it as a PR here? Note that I plan to go through and add a couple of extra features (which is mostly to say that I want to see if I can add additional helpful context data to the notifications) and would additionally be happy to make any other changes requested by you guys prior to submitting such a PR.

kyrylo:
> Would airbrake be willing to merge this implementation into their official gem/repository if I refactored it as a PR here? 

Hi, Mike! That would be fantastic. :+1:

mvastola:
Ok. Cool. I'm going to close this then. If not tomorrow (by which I mean Sunday), expect a PR soon!

mvastola:
@kyrylo, FYI, see #472. Input would be greatly appreciated!

PR #472, opened 2016-01-10
Added integration with Phusion Passenger v5.0.x

mvastola:
Following up on #470, here is what I have as a rough-draft. It's totally not ready for merging yet (apparently my original code greatly "offended" `rubocop`s style requirements), but I'd appreciate initial thoughts, both on below and the changes themselves.

Items on my TODO list are (to be edited/whittled down as I go along):
* Figure out what, if any, additional contextual info from `passenger` or elsewhere should be sent up to `airbrake`. (Perhaps a list of gems? Suggestions welcome.)
* CI Testing, of course. (Not quite sure how to set this up, actually..)
  * I'm used to `Travis` rather than `circle`, which is apparently what you have, but I wouldn't even be sure how to set it up on the former off the top of my head, so this may require some trial and error.
    * This is if testing is to include running a web server with a `passenger` module.
* "We currently use Airbrake::Error for every possible error this gem can raise."

### DIFF
--- a/lib/airbrake/passenger/integration.rb
+++ b/lib/airbrake/passenger/integration.rb
@@ -1,0 +1,53 @@
+module Airbrake
+  # Integration for Phusion Passenger
+  module Passenger
+    IntegrationError = Class.new(::StandardError)
+    class << self
+      def notifier
+        @notifier || :default
+      end
+
+      def notifier=(val)
+        is_sym = val.is_a?(Symbol)
+        @notifier = is_sym ? val.to_sym : val
+      end
+    end
+    # Implementation of Passenger Integration for Versions 4.0 thru 5.0
+    module Integration
+      VERSION_REQUIREMENT = Gem::Requirement.new('>= 4.0', '< 5.1').freeze
+      class << self
+        def integrate!
+          return @integrated if defined?(@integrated)
+          raise IntegrationError, "Your version of PhusionPassenger " \
+            "is not currently supported by the airbrake gem." unless compatible?
+          unless defined?(::PhusionPassenger::LoaderSharedHelpers)
+            ::PhusionPassenger.require_passenger_lib 'loader_shared_helpers'
+          end
+          ::PhusionPassenger::LoaderSharedHelpers.send :extend, ClassMethods
+          @integrated = true
+        rescue IntegrationError => e
+          Kernel.warn e
+        end
+
+        def compatible?
+          @compatible ||= VERSION_REQUIREMENT.satisfied_by?(passenger_version)
+        end
+
+        private
+
+        def passenger_version
+          @passenger_version ||= begin
+            unless defined?(::PhusionPassenger)
+              raise IntegrationError, "PhusionPassenger not detected."
+            end
+            Gem::Version.new(::PhusionPassenger::VERSION_STRING)
+          end
+        end
+      end
+
+      autoload :ClassMethods, 'airbrake/passenger/integration/class_methods'
+    end
+  end
+end
+
+::Airbrake::Passenger::Integration.integrate!
--- a/lib/airbrake/passenger/integration/class_methods.rb
+++ b/lib/airbrake/passenger/integration/class_methods.rb
@@ -1,0 +1,29 @@
+require 'airbrake/passenger/integration'
+module Airbrake
+  module Passenger
+    module Integration
+      # Module to be extended into ::PhusionPassenger::LoaderSharedHelpers
+      module ClassMethods
+        def self.extended(klass)
+          if klass.respond_to?(:about_to_abort)
+            klass.send :alias_method, :about_to_abort_without_notifier, :about_to_abort
+          end
+        end
+
+        def about_to_abort(*args)
+          options = args.first.is_a?(Hash) ? args.first.clone : {}
+          exception = args.last.clone if args.last.is_a?(Exception)
+          if exception
+            params = {}
+            params[:component] = "PhusionPassenger/" + ::PhusionPassenger::VERSION_STRING
+            params[:action] = options['process_title'] || 'Unknown'
+            # TODO: Add more contextual info here
+            ::Airbrake.notify_sync(exception, params, ::Airbrake::Passenger.notifier)
+          end
+          return unless respond_to?(:about_to_abort_without_notifier)
+          about_to_abort_without_notifier(*args)
+        end
+      end
+    end
+  end
+end
--- a/airbrake.gemspec
+++ b/airbrake.gemspec
@@ -40,4 +40,5 @@ DESC
   s.add_development_dependency 'webmock', '~> 1'
   s.add_development_dependency 'rack-test', '~> 0'
   s.add_development_dependency 'sidekiq', '~> 4'
+  s.add_development_dependency 'passenger', '>= 4', '<= 5.1'
 end
--- a/lib/airbrake.rb
+++ b/lib/airbrake.rb
@@ -10,6 +10,8 @@ require 'airbrake-ruby'
 require 'airbrake/version'
 
 # Automatically load needed files for the environment the library is running in.
+require 'airbrake/passenger/integration' if defined?(PhusionPassenger)
+
 if defined?(Rack)
   require 'airbrake/rack/user'
   require 'airbrake/rack/notice_builder'


mvastola:
So I am now correctly getting `airbrake` notifications on the latest `passenger` enterprise if I put the following in a Gemfile:
```
gem 'airbrake', '~> 5.0.0', :github => 'Partyista/airbrake'
```
And adding the following on top of (or replacing entirely) the `config.ru` (with real values in the configuration):
```
gem 'airbrake'
require 'airbrake'
Airbrake.configure do |c|
  c.project_id = 123456 
  c.project_key = 'abcde1234567890' 
end
raise 'Fake Exception!'
```

After handling the exception by triggering the `airbrake` notifier, the exception is propagated back to `passenger` which handles it in accordance with your configuration (either a fancy error page or a web server 500 page) the same way it would without `airbrake`.

kyrylo:
Hey, thanks for the PR! I tried to test this with a sample Rails app on Heroku, but I don't get any exceptions. Here's my test app: https://github.com/kyrylo/passenger-airbrake

```
2016-01-11T09:07:00.808301+00:00 app[web.1]: App 135 stdout: 
2016-01-11T09:07:00.819076+00:00 app[web.1]: [ 2016-01-11 09:07:00.8189 65/7f76c99ec700 age/Cor/App/Implementation.cpp:304 ]: Could not spawn process for application /app: An error occurred while starting up the preloader.
2016-01-11T09:07:00.819116+00:00 app[web.1]:   Error ID: 02b7bba4
2016-01-11T09:07:00.819138+00:00 app[web.1]:   Error details saved to: /tmp/passenger-error-3GQKVz.html
2016-01-11T09:07:00.819157+00:00 app[web.1]:   Message from application: hell (RuntimeError)
2016-01-11T09:07:00.819175+00:00 app[web.1]:   config.ru:13:in `block in <main>'
2016-01-11T09:07:00.819194+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/rack-1.6.0/lib/rack/builder.rb:55:in `instance_eval'
2016-01-11T09:07:00.819212+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/rack-1.6.0/lib/rack/builder.rb:55:in `initialize'
2016-01-11T09:07:00.819230+00:00 app[web.1]:   config.ru:1:in `new'
2016-01-11T09:07:00.819268+00:00 app[web.1]:   config.ru:1:in `<main>'
2016-01-11T09:07:00.819284+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:110:in `eval'
2016-01-11T09:07:00.819302+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:110:in `preload_app'
2016-01-11T09:07:00.819320+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:156:in `<module:App>'
2016-01-11T09:07:00.819343+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:30:in `<module:PhusionPassenger>'
2016-01-11T09:07:00.819362+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:29:in `<main>'
2016-01-11T09:07:00.819378+00:00 app[web.1]: 
2016-01-11T09:07:00.819395+00:00 app[web.1]: 
2016-01-11T09:07:00.826847+00:00 app[web.1]: [ 2016-01-11 09:07:00.8251 65/7f76c1ffd700 age/Cor/Con/CheckoutSession.cpp:277 ]: [Client 2-1] Cannot checkout session because a spawning error occurred. The identifier of the error is 02b7bba4. Please see earlier logs for details about the error.
```

Any ideas how to make it report? Perhaps, there are other easier ways that can help me to test this?

mvastola:
Hmm.. well one aspect of this patch (which was supposed to be a plus) was that it's almost totally transparent/silent. Though you should have gotten warnings if you `require`d `airbrake/passenger` directly and my PR wasn't able to patch the correct functions.. will have to try to debug on my own heroku app.

One thing I just saw though is that you have a `gem 'passenger'` line. Try removing that.  Per their docs it shouldn't be in your `Gemfile` I don't believe (and if you need to put it there for heroku, at least add a `:require => false`. Also try removing `require 'airbrake/passenger'`. There definitely could be an issue with `require`ing my code twice. If either of those places are redefining the methods I monkey-patched, that would neuter the integration. (I'll have to put some sort of guard or warning in to catch that.)

kyrylo:
Thanks for the info. I updated the app. It was the spurious `require` that I forgot to delete. I also added `require: false` as per your suggestion. I was able to make it work, but only one time. Subsequent deploys don't send exceptions to the Airbrake dashboard anymore :(

I'd be really grateful if you could fork or create a repo that is reliably sending errors to the dashboard. Then I could properly review your PR :)

mvastola:
Hmm. Well that's rather odd, and of course I'll look at it. One thought though: can you set `airbrake` to send synchronous notifications in your config? I wonder if passenger could be terminating the thread (when it terminates itself) that sends the notification before it can get sent out since you're doing those asynchronously now, IIRC.

kyrylo:
Airbrake falls back to synchronous mode when async send can't be performed. That said, to be 100% sure you could replace `notify` with [`notify_sync`](https://github.com/airbrake/airbrake-ruby#airbrakenotify_sync) in your repo (no more `config.async`). Once we have that, I could try again.

mvastola:
Ok. I'll use `notify_sync`. What I'm saying is if the web server itself is quitting, it might take the `airbrake` thread with it at the process level, regardless of a fallback.

mvastola:
So without changing anything in my PR (other than rebasing with two recent commits to this repo), I created https://github.com/Partyista/airbrake-passenger-demo, which is a greatly simplified version of your repo, and I hosted it on heroku at secret-tundra-7554.herokuapp.com with all default settings.

It seems to be working just fine (and I haven't even switched to `notify_sync` yet). 
```
2016-01-12T02:47:15.402563+00:00 app[web.1]: App 167 stdout: 
2016-01-12T02:47:15.776887+00:00 app[web.1]: App 167 stdout: D, [2016-01-12T02:47:15.776647 #167] DEBUG -- : **Airbrake: {"id"=>"1596040346155152784", "url"=>"https://airbrake.io/locate/1596040346155152784"}
2016-01-12T02:47:15.819567+00:00 app[web.1]: [ 2016-01-12 02:47:15.8194 65/7f67e24d6700 age/Cor/App/Implementation.cpp:304 ]: Could not spawn process for application /app: An error occurred while starting up the preloader.
2016-01-12T02:47:15.819571+00:00 app[web.1]:   Error ID: f82a6267
2016-01-12T02:47:15.819582+00:00 app[web.1]:   Error details saved to: /tmp/passenger-error-9ZrzK5.html
2016-01-12T02:47:15.819584+00:00 app[web.1]:   config.ru:9:in `block in <main>'
2016-01-12T02:47:15.819585+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/rack-1.6.4/lib/rack/builder.rb:55:in `instance_eval'
2016-01-12T02:47:15.819586+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/rack-1.6.4/lib/rack/builder.rb:55:in `initialize'
2016-01-12T02:47:15.819587+00:00 app[web.1]:   config.ru:1:in `<main>'
2016-01-12T02:47:15.819583+00:00 app[web.1]:   Message from application: Fake Exception! (RuntimeError)
2016-01-12T02:47:15.819593+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:29:in `<main>'
2016-01-12T02:47:15.819588+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:110:in `eval'
2016-01-12T02:47:15.819587+00:00 app[web.1]:   config.ru:1:in `new'
2016-01-12T02:47:15.819589+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:110:in `preload_app'
2016-01-12T02:47:15.819591+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:30:in `<module:PhusionPassenger>'
2016-01-12T02:47:15.819590+00:00 app[web.1]:   /app/vendor/bundle/ruby/2.0.0/gems/passenger-5.0.23/src/helper-scripts/rack-preloader.rb:156:in `<module:App>'
2016-01-12T02:47:15.819594+00:00 app[web.1]: 
2016-01-12T02:47:15.819595+00:00 app[web.1]: 
2016-01-12T02:47:15.825005+00:00 app[web.1]: [ 2016-01-12 02:47:15.8247 65/7f67bdffd700 age/Cor/Con/CheckoutSession.cpp:277 ]: [Client 5-1] Cannot checkout session because a spawning error occurred. The identifier of the error is f82a6267. Please see earlier logs for details about the error.
2016-01-12T02:47:15.809213+00:00 app[web.1]: App 167 stdout: 
2016-01-12T02:47:15.825739+00:00 heroku[router]: at=info method=GET path="/" host=secret-tundra-7554.herokuapp.com request_id=f87c03a0-9b14-44a8-92b6-2e4b72d085c0 fwd="[IP REDACTED]" dyno=web.1 connect=2ms service=484ms status=500 bytes=2847
```
(Note I'm sending the notifications to your Airbrake account because I'm on the most basic plan that only supports one project, so you'll have to confirm you're getting them correctly, but the logs seem to suggest the notifications are being created, which is a huge step up..)

So I'm at a loss for why yours might have been failing, but I'll check a few things.. I'm happy to give you access to my demo heroku/github repos if you'd like to futz around with things. Thoughts on how to go from here are welcome. In the interim there are a few aspects of my PR I want to refactor anyway.

mvastola:
Also, @kyrylo, if it'd be helpful to do any of this off GitHub via email or IM, feel free to get in touch via my email in my GitHub profile. (I have a feeling some amount of collaboration might be needed to track down what's causing us to get different results with what -- ostensibly -- is the same code.)

kyrylo:
Thanks! I managed to get it working reliably only by replacing `notify` with `notify_sync`.

lib/airbrake/passenger.rb
```diff
@@ -0,0 +1,62 @@
+require 'set'
+require 'pathname'
+
+module Airbrake
+  # Integration for Phusion Passenger
+  module Passenger
+    IntegrationException = Class.new(::Exception)
```
kyrylo: We currently use `Airbrake::Error` for every possible error this gem can raise.

lib/airbrake/passenger.rb
```diff
@@ -0,0 +1,62 @@
...
+      else
+        @integrated = true
+      end
+
+      def integrated?
+        @integrated == true
+      end
+    end
+  end
+end
```
kyrylo: The convention we try to follow is to place every integration file inside its own directory. It would be nice if you could also put this file inside the passenger directory.

lib/airbrake/passenger/integration1.rb
```diff
@@ -0,0 +1,40 @@
...
+        def self.extended(klass)
+          if klass.respond_to?(:about_to_abort)
+            klass.send :alias_method, :about_to_abort_without_notifier, :about_to_abort
+          end
+        end
+
+        def about_to_abort(*args)
+          exception = args.last if args.last.is_a?(Exception)
+          if exception
+            notice = ::Airbrake.build_notice(exception)
```
kyrylo: You don't have to `build` notices here because you can pass them to `notify` [directly](https://github.com/airbrake/airbrake/blob/master/lib/airbrake/resque/failure.rb) like shown.

lib/airbrake/passenger.rb
```diff
@@ -0,0 +1,62 @@
...
+        return @integrated if defined?(@integrated)
+        integration.integrate!
+      rescue IntegrationException => e
+        Kernel.warn e
+        @integrated = false
+      else
+        @integrated = true
+      end
+
+      def integrated?
```
kyrylo: This method seems to be unused.

lib/airbrake/passenger.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      def integrations
+        return @integrations if defined?(@integrations)
+        @integrations ||= Set.new
+        dir = Pathname.new(__FILE__).dirname.join('passenger')
+        dir.each_child do |child|
+          next unless child.file? && child.readable?
+          next unless child.fnmatch?('*.rb')
+          require "airbrake/passenger/#{child.basename '.rb'}"
```
kyrylo: I feel like we can simplify this code a lot by not doing these shenanigans. Why do you need to keep track of "integrations" in the first place?

lib/airbrake/passenger.rb
```diff
@@ -0,0 +1,62 @@
...
+    IntegrationException = Class.new(::Exception)
+    class << self
+      def version
+        unless defined?(::PhusionPassenger)
+          raise IntegrationException, "PhusionPassenger not detected"
+        end
+        @version = Gem::Version.new(::PhusionPassenger::VERSION_STRING)
+      end
+
+      def version?(req)
```
kyrylo: This method name doesn't tell a lot to me.

kyrylo:
At the moment it seems like we're using a sledgehammer to crack a nut. Do you think you could simplify this by leaving only the necessary code? That is, I don't think there's need to version our integrations for now.

kyrylo:
Addressing your first bullet: considering that these errors can only occur very early, you could configure multiple notifiers and write this integration accordingly.

```ruby
# config.ru
Airbrake.configure(:passenger) do |c| 
  c.project_id = xxx 
  c.project_key = 'xxx'
end

# our integration
Airbrake.notify_sync(ex, params, :passenger)

# config/initializers/airbrake.rb
Airbrake.configure do |c| 
  c.project_id = xxx 
  c.project_key = 'xxx'
end
```

Happy to discuss this because as you can see I'm not extremely familiar with Passenger.

mvastola:
> At the moment it seems like we're using a sledgehammer to crack a nut. Do you think you could simplify this by leaving only the necessary code? That is, I don't think there's need to version our integrations for now.

There are a couple of answers to this.
  * The simple one is "Yes." If you really want me to, obviously it can be done, and I would do it for the sake of getting this PR merged, but I would disagree that it is optimal. (If anything, I would suggest moving this code -- it can be made sufficiently robust quite easily I would think -- up one layer and using it throughout this gem to determine how to integrate `airbrake` into different products/versions.)
  * Since, as you mention, you aren't to familiar with passenger, I should probably explain that passenger is a bit weird in how it works with this gem and how this gem must interact with it. In short, there's no way to version it within an app or within a `Gemfile`, since it's not even supposed to be in the `Gemfile`. To run it, you are just supposed to:
    * For the standalone version, do: `gem install passenger` and then run `passenger` with the correct arguments.
    * For the Apache/Nginx versions: `gem install passenger`, run `passenger-install-{apache2,nginx}-module` (or skip those two and install the package), configure Apache/Nginx and run it.
  
  After doing so, despite `passegner` not being included as a gem, the module `::PhusionPassenger` will be defined/accessible from inside the ruby app because `passenger` bootstraps in its own set of helper libraries prior to running `config.ru`.  In other words, there is no way to enforce compatibility with specific versions of `passenger` via the standard `RubyGems`/`Bundler`/`Gemfile`/`gemspec` setup, so the reality is there is no way to stop this code from running with *any* version of `passenger`. The best thing we can do is to print a warning to the logs (and actually -- we should send a specific airbrake notification.. I just thought of this) if we can't handle the version being run. The rest of the functionality provides robustness for future expansion.
  * Though you haven't run into this yet it would seem, this sort of integration code (and, admittedly, complexity) is relevant across all gems you might want to integrate with. They will come out with newer versions that will need to be integrated with differently based on the version of the other gem. This code makes things much more modular and ensures that the correct integration is used in all situations.

mvastola:
Re: 
> Addressing your first bullet: considering that these errors can only occur very early, you could configure multiple notifiers and write this integration accordingly.
Yea, I'm aware of that ability. It just seems somewhat superfluous to me to configure a second notifier with the same `#project_id` and `#project_key` as the first. And in some instances the notifiers will be identical so users might not want to be bothered defining them twice. I suppose I could create a `Airbrake::Passenger.notifier` accessor that defaults to `:default` but otherwise let users configure it to any notifier they wish. (Might not be a bad feature for all integrationons.)

lib/airbrake/passenger.rb
mvastola: Yea. Will change.

lib/airbrake/passenger.rb
mvastola: Not a problem.

lib/airbrake/passenger.rb
mvastola: This was one of those things I was coding quickly and not really looking to simplify code. I also did this for a particular purpose. `IntegrationExceptions` in `Airbrake::Passenger` are exceptions that I want to `raise` to cancel the integration but then be caught and issued as a warning for the logs so as not to kill the server booting. 

In any case, I'm sure there's a neater way to do this in keeping with your current practices.

lib/airbrake/passenger/integration1.rb
mvastola: I'll look at the docs on how this works. I really hadn't the first time. I forget where I copied this code from. (I think another integration.)

lib/airbrake/passenger.rb
mvastola: I think I expected to use it for something. If it goes unused I'll trash it.

lib/airbrake/passenger.rb
kyrylo: I hadn't seen that you catch it within the same file. I am not sure why we need to raise it in the first place.

lib/airbrake/passenger.rb
mvastola: I'm planning to do some majoring refactoring to this. I agree that this method is a bit unwieldy. I think it's use will be more clear when its more concise. The idea is to iterate through the `#iterations` to find one with version constraints that match the version of `passenger` in use, and then to use that integration.

lib/airbrake/passenger.rb
mvastola: It's entirely possible there's a better way to do it, but it's raised on lines 11 and 40 and caught in the '#integrate!` method.

mvastola: Line 48.

mvastola:
> Thanks! I managed to get it working reliably only by replacing `notify` with `notify_sync`.

Well that's odd. No idea why I wasn't having the same issue. Maybe it was a time of day thing and Heroku was more/less overloaded when you were running it (or on the the particular server you were using)?

In any case, I'll switch my code to synchronous, since that seems to be key.

mvastola:
PS: Just disabled that heroku deployment for the time being. I think it proved its usefulness and I don't want to overload you with notifications (I don't know how many web crawlers github gets).

lib/airbrake/passenger.rb
mvastola: Fixed.

lib/airbrake/passenger.rb
mvastola: Fixed. (Somewhat)

lib/airbrake/passenger.rb
mvastola: Removed.

lib/airbrake/passenger/integration1.rb
mvastola: Added to `TODO`.

lib/airbrake/passenger.rb
mvastola: Added to `TODO`.

lib/airbrake/passenger.rb
mvastola: Done.

mvastola:
Just pushed through a few changes and replied to your inline comments. Added some more things to do onto my `TODO` list and waiting for followups from you to figure out where to go on the rest.

kyrylo:
Thanks for the Passenger intro, it's helpful :)

> If anything, I would suggest moving this code -- it can be made sufficiently robust quite easily I would think -- up one layer and using it throughout this gem to determine how to integrate airbrake into different products/versions

I agree with this approach from the point of theory, but I disagree with this PR, which has this mechanism only for 1 integration. I also don't see the strong need to have this abstraction at the moment. Designing this mechanism is no easy task, because every integration integrates differently. So far simple `if` statements can deal different versions just fine. The code might not be the prettiest, but it is very easy to support and understand so far.

> This code makes things much more modular and ensures that the correct integration is used in all situations.

Reality shows that the current approach works. Please don't think that I am totally against your approach, but *at the moment* I don't want to solve the problem, which doesn't exist yet.

> In any case, I'll switch my code to synchronous, since that seems to be key.

I'll try to play with it a little bit more after we're settled with the final code, just to be 101% sure. It will be easy to switch back and forth.

>  It just seems somewhat superfluous to me to configure a second notifier with the same #project_id and #project_key as the first

I tend to agree with this, but as you previously mentioned, requiring the initializer file isn't the best approach. We also cannot simply move the initializer to `config.ru` and recommend every Passenger user use that. Why? Simply because the initializer config configures some useful options like `environment`, which can't be obtained in `config.ru`, can they? So second notifier would lack some options.

mvastola:
*Re: Integration Versioning*
All fair points. I would suggest perhaps that the reason you don't *need* this layer of abstraction *yet* (and the reason the current approach has worked thus far) is because none of the gems you're supporting integrating with have changed major versions while the current iteration of the `airbrake` gem has been around. (I suspect your first taste of this may actually be Rails 5, which is in beta now, but I haven't really looked too closely at the Rails integration to be sure.) 

Anywho, I'll rip that code out and do a more major refactor. I'll branch my code with this feature though in case it has a future use.  

*Re: Synchronicity*
I suspect my initial guess is correct and the aborting on the web server process is bringing down the airbrake notification thread before it sends. Going synchronous should definitely solve the problem. It's just odd to me how we both didn't experience this race condition.

*Re: Notifiers*
Thinking more about it, I think we're forced to do it with a separate notifier (I like the idea of making its name configurable), because -- while incredibly unlikely -- the `passenger` notifier can trigger at any point during page load if `rack` experiences an exception in its own code that it doesn't catch. (I was originally going to reiterate my earlier suggestion to change `airbrake-ruby` to allow for the amendment/redefinition of notifier configurations, but that clearly can't work since we need to allow for this edge case.)

kyrylo:
> I suspect your first taste of this may actually be Rails 5, which is in beta now, but I haven't really looked too closely at the Rails integration to be sure.

We automatically test our Rack integration against Rails 5.

> Anywho, I'll rip that code out and do a more major refactor. I'll branch my code with this feature though in case it has a future use.

Sounds reasonable.

> I was originally going to reiterate my earlier suggestion to change airbrake-ruby to allow for the amendment/redefinition of notifier configurations

I wouldn't want to do that because that will make debugging the gem much harder. In addition, there's no real need to change configuration options: they're all static.

<hr> 

From my understanding this integration will be catching exceptions *only* during a short period of application's startup process. It won't be catching Passenger exceptions once it's deployed because there will be none. If that's true, then multiple notifiers looks like a goodish solution to me.

mvastola:
> We automatically test our Rack integration against Rails 5.

Oh, didn't know that. Cool.

> From my understanding this integration will be catching exceptions only during a short period of application's startup process. It won't be catching Passenger exceptions once it's deployed because there will be none. If that's true, then multiple notifiers looks like a goodish solution to me.

I'd have to look at the code more, but I think technically this integration *could* catch an exception at any time, it's just unlikely for it to do so once `rack` initializes because `rack` is designed to handle any/all exceptions that occur after that point. If one did slip through, however, and it passed through to `passenger`, this integration should, in fact, handle it. 

It's also worth noting that `passenger` doesn't strictly *require* that `rack` be used. So it's possible that this could be the *only* integration present in some really irregular home-grown setup. It would still work though.

In any case, this (albeit very unlikely) scenario rules out the possibility of amending notifier configurations. I was thinking that you could set the `#project_key` and `#project_id` in `config.ru` and then add more options (such as the environment name) to the same named `Notifier` in a `rails` initializer once `rails` loads, but that wouldn't work since if an error occurs in `rack` *after* `rails` has loaded, you still don't want the additional options to the `Notifier`.

mvastola:
> We automatically test our Rack integration against Rails 5.

Oh, I see you already have this to make Rails 5 work:
```
if ::Rails.version =~ /\A5\./
```

mvastola:
Ok. Doing some more changes now. Just committed the removal of that layer of abstraction with the integration versions. If you ask me, the code looks more unwieldy and harder to read now, but it's up to you. I'm going to keep working on the rest of the changes for now.

mvastola:
@kyrylo See current commit. Things are almost good to go. I just need to figure out testing (which I should be able to do on my own), and I wanted to talk to you about `Airbrake:Error`.. 

It seems that that's defined in the `airbrake-ruby` package and that those errors are used to indicate problems within that gem, mostly surrounding configuration and sending error notifications. The exceptions I was raising aren't at all similar, I wouldn't say. I also noticed at least one other integration that raised and caught errors without going through `Airbrake::Error` (I forget.. I will have to double-check which). Let me know how (if at all) you advise I rework the way my code handles the two `IntegrationErrors` it uses as of now.

Thanks

lib/airbrake/passenger/integration/class_methods.rb
```diff
@@ -0,0 +1,28 @@
...
+            klass.send :alias_method, :about_to_abort_without_notifier, :about_to_abort
+          end
+        end
+
+        def about_to_abort(*args)
+          exception = args.last if args.last.is_a?(Exception)
+          if exception
+            params = {}
+            params[:component] = "PhusionPassenger " \
+              "v#{::PhusionPassenger::VERSION_STRING}"
```
vmihailenco: @kyrylo probably this should be appended to `context[:version]`, e.g. `1.2.3 Rails/4.1.6 PhusionPassenger/1.1.1`.

lib/airbrake/passenger/integration.rb
```diff
@@ -0,0 +1,59 @@
...
+          return @integrated if defined?(@integrated)
+          raise IntegrationError, "Your version of PhusionPassenger " \
+            "is not currently supported by the airbrake gem." unless compatible?
+          unless defined?(::PhusionPassenger::LoaderSharedHelpers)
+            ::PhusionPassenger.require_passenger_lib 'loader_shared_helpers'
+          end
+          ::PhusionPassenger::LoaderSharedHelpers.send :extend, ClassMethods
+        rescue IntegrationError => e
+          Kernel.warn e
+          ::Airbrake.notify_sync(e, {}, notifier)
```
vmihailenco: I doubt notifier is configured at this point...

mvastola: Ohh.. true. Wasn't thinking. Guess I'll need to remove that line.

lib/airbrake/passenger/integration/class_methods.rb
mvastola: Can do.. let me know what is preferred.

lib/airbrake/passenger/integration.rb
mvastola: @kyrylo Hmm.. is it OK to check and see if they `require`d `airbrake` manually here (as opposed to `bundler` doing it) and if so raise the exception instead of just warning it?

Because if someone has required `airbrake` prior to the `bundler` invocation, it's all but guaranteed they're trying to set up `airbrake` integration so that it can catch all exceptions. Throwing a hard error will at least make it obvious something went wrong, and `passenger` (in dev mode at least) will display a friendly error page with a backtrace. Otherwise it's at least a huge hint to check the trace in the logs.

mvastola: Alternatively, you could argue it makes sense to put *all* integrations as a callback after the first `Airbrake.configure` block gets run to facilitate this sort of error notification..

mvastola: Or you could `require 'airbrake-ruby'`, do `Airbrake.configure` and then `require 'airbrake'`.. (though in this case you have no opportunity to configure the `#notifier` accessor.)

..so several posibilities.

kyrylo:
Hey! I've made a simplified version of this PR: https://github.com/airbrake/airbrake/pull/479

Massive thanks for your help (:heart: :yellow_heart: :green_heart: ) and the initial implementation, but I felt it was quicker to roll out my own.

Interestingly enough, it "just works". You don't need to configure anything at all. Could you please confirm that it works for you?

mvastola:
Cool. I'll check tonight. I'm worried about a few potential issues off the top of my head..

No idea whatsoever how you're getting by without configuring it separately.. I'm concerned it doesn't catch all errors in that case.. but I'll check.

mvastola:
@kyrylo I have a rough outline of integration tests that could be configured, but I'm not sure if I'm going overboard or doing this the wrong way, etc, so I'd like more guidance on the tests before I continue.

I also updated the first commit so that it's (AFAIK) ready to merge and resolves every issue that has been brought up.

Appraisals
```diff
@@ -6,7 +6,7 @@ appraise 'rails-3.2' do
   gem 'sqlite3', '~> 1.3.11', platforms: [:mri, :rbx]
 
   gem 'resque', '~> 1.25.2'
-  gem 'resque_spec', git: 'git@github.com:kyrylo/resque_spec.git'
+  gem 'resque_spec', github: 'kyrylo/resque_spec'
```
mvastola: FYI, I updated these because they are currently in SSH format, so `bundler` is trying to SSH-authenticate prior to fetching this git repo.

mvastola:
Update: I have a bunch of tests running and passing now, though the setup is looking extremely ugly still.

kyrylo:
Hey, Mike!

Thanks for your work on this PR. I can see it has become a lot more complex with the tests. I can see that you monkey-patch standard Ruby classes. This would be pretty hard to support. We've talked about your integration and came to conclusion that the best way to move forward would be:
* deleting the test suite and writing a good description in the comments about what exact errors this integration is expected to catch
* leaving only the necessary code to catch exceptions (class_methods.rb)
* documenting this integration in the README and showing examples of how to integrate it with Rails/Rack apps

mvastola:
Interesting idea.. I wouldn't have thought of that. So like people would have to add `PhusionPassenger::LoaderSharedHelper.extend (...)` manually? And then we just document the known-compatible versions as opposed to checking for them?

If I'm understanding this correctly, this seems like an excellent compromise if you don't feel up for maintaining the other half of the patch. 

I agree the test suite in its current form is ultra unweildy. Something certainly needed to be done with it. I was in the process of trying to figure out what, but I don't have a problem just ditching it entirely if it ceases to be relevant?

Please just tell me exactly what you want me to do on my end and I'll do it.

mvastola:
Seeing as 190ee8d was pushed today (which I'm happy to see), I guess I should close this. I'm honestly *a little* torn as I think some simplifications sacrificed useful features/options/checks that didn't add too much bloat IMHO, but I appreciate you spending time working with me nonetheless to listen and try to address as many of my concerns as possible.

kyrylo:
I simply rebased that branch against master because I needed it for some other unrelated to the integration and this PR testings.

mvastola:
Ohh. Sorry. Fail. Looking at this on my phone and all I got was an email commit notification about 190ee8d. Just assumed it was to the master branch. (It's too early in the morning!) So I assumed you just decided to ignore my question and go with your own way.

I hope we still get a chance to discuss before the commit because there are several things I'd change there (although I've mentioned most of them by now I'm sure).

mvastola:
I've never been in a situation where the owners of a gem rolled their own competing PR in response to my PR, so I guess I forgot the branch for the other PR was in this repo.

kyrylo:
It's not competing. I published that in order to show you my thinking and simplify explanations. While working on that I was also able to better understand how Passenger works.

As for this PR and your questions, I need more time to answer them.

mvastola:
@kyrylo, alright. I interpreted your comments here and the commit message as you were going to take the your new PR and merge that instead. No matter though..

I just updated the code in this PR, to both rebase and remove the integration tests. I think at this point it might be helpful to compare and contrast the approaches between the two PRs. That way we can break things down as much as possible and pick and chose which aspects of which PRs are ideal in order to resolve this.

mvastola:
## Features in #472 not in #479
* Ability to specify notifier name (required if notifiers triggered later on will include extra info)
* Notify for errors in `Gemfile` and/or before platform (i.e. `rack`) is fully loaded/configured.
* Protection against double-notifying if the user tries to include the library files manually

That's all I got right now. I'll have to look things over a bit more though.

mvastola:
FYI, I fixed a `rubocop` complaint (inconsequential) and was causing the integration check to fail.

lib/airbrake/passenger/integration.rb
```diff
@@ -0,0 +1,53 @@
...
+          @passenger_version ||= begin
+            unless defined?(::PhusionPassenger)
+              raise IntegrationError, "PhusionPassenger not detected."
+            end
+            Gem::Version.new(::PhusionPassenger::VERSION_STRING)
+          end
+        end
+      end
+
+      autoload :ClassMethods, 'airbrake/passenger/integration/class_methods'
```
kyrylo: Why do you need autoload here?

kyrylo:
Hey, Mike!

After some more thoughtful consideration about the Passenger integration we came to conclusion that the best way to advance here would be having a separate gem. We would like to see how it fares first. Given your deep knowledge about Passenger, we think you will be the best maintainer for this code. We could link to your gem from our README. If it gets popular enough we can surely talk about merging it with the Airbrake gem again.

Cheers!

lib/airbrake/passenger/integration.rb
mvastola: Hm. I don't. It's a holdover from when this had the abstraction layer for handling integrations into different versions of the same gem differently. It could be replaced with a require at the top.

mvastola:
Alright. If you're not interested in adding this functionality then that is fine. I will have to figure out if we want to launch a gem though. We may end up just hard-coding this integration into our app and leaving it there.. I'll let you know either way.

kyrylo:
That's also fine, thanks for letting me know! We have this PR as a reference, so if we'll see enough user requests, we can always get back to this.

mvastola:
Cool. The code is yours for the taking!

PR #475, opened 2016-01-12
capistrano: fix the Capistrano 2 integration

### LINKED ISSUE #468: Capistrano 2.x deployment notification broken
```
fileutils.rb:127:in `chdir': No such file or directory @ dir_chdir
```
I'm not sure how this would work, this command would run on the local machine doing the deployment:
https://github.com/airbrake/airbrake/blob/master/lib/airbrake/capistrano/tasks.rb#L42

It needs to change the directory on the remote machine.

kyrylo:
Fixes #468 (Capistrano 2.x deployment notification broken)

The command we're running here should be run on the remote machine
(only once). The previous code was running it locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Airbrake Changelog
   ([#469](https://github.com/airbrake/airbrake/pull/469))
 * Added better support for user reporting for Rails applications (including
   OmniAuth support) ([#466](https://github.com/airbrake/airbrake/pull/466))
+* Fixed the Capistrano 2 integration, which was not working at all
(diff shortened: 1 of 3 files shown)


vmihailenco:
LGTM

nick-desteffen:
The previous integration would change to the release directory:
https://github.com/airbrake/airbrake/blob/airbrake-v4/lib/airbrake/capistrano.rb#L79-L80

You may need to do this in order for it to work.

kyrylo:
Good catch. That's actually what I wanted to do but for some reason it didn't go into the commit. Fixed this and tested locally. We should be good here now. Thanks!

PR #479, opened 2016-01-14
Implement the Passenger integration

kyrylo:
This implementation wouldn't be possible without @mvastola's help.
This PR is based on [his work][1] and offers a simpler integration.

I don't think there's a straightforward way to test this integration,
so I'm leaving it without tests.

[1]: https://github.com/airbrake/airbrake/pull/472


mvastola:
**WORK IN PROGRESS (I'll remove this line when this comment is complete in a few hours)**

Ok. So issues I see with this:
* If, `passenger` is changed in some way that is incompatible with this implementation and `passenger` is upgraded on a host (keeping in mind `passenger` can be upgraded through package upgrades or by web hosts, usually entirely separately from the `Gemfile`), whilst `airbrake` is not, the lack of a conditional `extend`/`include` contingent on `passenger`'s version could theoretically bring down an entire site.
* I don't think this code will work with `passenger` 4.0.x (whereas mine did). In that version, `#about_to_abort` only took one parameter: the exception, and this parameter was optional.
* On [this](https://github.com/airbrake/airbrake/pull/479/files#diff-25fdca1bcc6f1c577c74b1c7f4e63486R14) line you have `params = args.first`. The first argument is named `options`, but I'm not 100% sure it has the values you're implying. If it does, I'm afraid it may also have values that are security-sensitive and need to be eliminated prior to transmission. Also this argument isn't present in 4.0.x and (I believe) is optional in 5.0.x, so you will need to check for its existence.
* Subdirectory name should be changed to `phusion_passenger` if the module name is going to be `PhusionPassenger`.
* Alternate notifier: how can this integration not have an option to use an alternate notifier configuration? As you yourself pointed out, there will be things like the Rails environment which will have not loaded yet at certain points. There would need to be a separate notifier configuration or else `airbrake` would call, for instance `Rails.env` in the middle of processing an error notification and try to initialize `rails` just to get the environment name in the middle of trying to perform an action on an already broken setup.
* TODO: check if multiple `require`s causes double notifications on error.
* TODO: Raise an error before `bundler` would kick in (or in the `Gemfile`) and see if it's caught.
* TODO: See how configuration is magically working.
* TODO: Testing. I think we can get something working.

In general, I respect that you're going for lightweight/simplicity, but I'm not 100% sure that's ideal in this type of gem from a design standpoint. As a user, I would really value robustness and plugability in this gem so that I could expand it to my custom needs. Nevertheless, it's your gem, so it's your call. This experience just has me debating if I should simply use `airbrake-ruby` and roll my own solution beyond that.

kyrylo:
I just pushed an update that addresses the issues you mentioned. Thanks!

> If, passenger is changed in some way ...

I added version restrictions.

> I don't think this code will work with passenger 4.0.x 

You were correct, fixed.

> On this line you have params = args.first. The first argument is named options, but I'm not 100% sure it has the values you're implying

It has at least in 5.0.23 and I modified the code so it has fallback values.

> I'm afraid it may also have values that are security-sensitive and need to be eliminated prior to transmission

Use the Airbrake.add_filter API to filter them out.

> Alternate notifier: how can this integration not have an option to use an alternate notifier configuration? As you yourself pointed out, there will be things like the Rails environment which will have not loaded yet at certain points.

Ideally this integration should work not only for Rails apps. In Rails apps it works with this `config.ru`:

```ruby

require ::File.expand_path('../config/environment', __FILE__)
raise 'asdasdasd'
run Rails.application
```

The env is loaded, so is our integration.

It doesn't work in this scenario:

```ruby
raise 'asdasdasd'
require ::File.expand_path('../config/environment', __FILE__)
run Rails.application
```

> check if multiple requires causes double notifications on error.

There's no need to `require`

> Raise an error before bundler would kick in (or in the Gemfile) and see if it's caught.

It wouldn't, but Passenger won't start in this case:

```
% bundle exec passenger start

[!] There was an error parsing `Gemfile`: asdsadsadas. Bundler cannot continue.

 #  from /home/kyrylo/code/passenger-ruby-rails-demo/Gemfile:5
 #  -------------------------------------------
 #  
 >  raise 'asdsadsadas'
 #  
 #  -------------------------------------------
```

> Testing. I think we can get something working.

Suggestions are welcome.

<hr>

I'd like to ask what errors are you expecting to catch exactly? It can already catch exceptions about missing gems, but what else?

mvastola:
Can't address everything right now, but:
* you wouldn't run `bundle exec passenger`. Just `passenger`, since passenger isn't in the `Gemfile`. In this case the error doesn't come up as soon.
* if it doesn't work if the error is raised before the initializer is loaded, that means it's not necessarily raised if there is a typo in a rails config file or another initializer.
* If `passenger` puts sensitive info ALWAYS into the `params` variable, I would think the info should be filtered out automatically without it needing to be done manually..

kyrylo:
> If passenger puts sensitive info ALWAYS into the params variable, I would think the info should be filtered out automatically without it needing to be done manually..

There's no way to "unfilter", so it's your decision as a user what you don't want to see.

mvastola:
True, but it's never a good idea to send a password that can be used to debug the instance (and thus get a full-functioning `irb` console) as would be in the case of enterprise `passenger`, over plaintext email, so no one should ever unfilter it..

mvastola:
> Suggestions are welcome.

https://gist.github.com/60e490aa83a6b3735b73

This gist is an all-inclusive ruby class (called `PassengerApp`) that will run the passenger server (via the `#run!` instance method) against a dynamically provided `config.ru` file (on initialization), makes an HTTP request to the root path on that server, shuts down the server (normally if possible, but forcefully if that fails) and then makes the server's log output available (via the `#log` instance method). 

I have this one actually make use of the gem, so it can be version locked and actually used with `Appraisals` or something similar with CI infrastructure. (Right now I have `passenger` specified via an inline `Gemfile`, but that was just so I could put everything into a single-file gist.)

Note this hasn't been tested with `passenger` 4.0.x yet, and it's possible some command line arguments differ (or aren't supported) there, but other than that, this should be good to go.

mvastola:
I'm still concerned about the other problems I listed here that you haven't addressed though. It seems like to properly/fully address those, we're back to my PR #472 again. (Even *without* addressing those, this code is looking more and more like the most recent version of #472..) 

I'm also not sure how I like (or if it's appropriate to) enclose the entire `module` in a conditional and/or have the completely unscoped `*_vsn` variables declared with toplevel binding (though I realize they're out of scope in any other files.. it just looks ugly to me, even if it doesn't offend `rubocop`).

kyrylo:
Closed because of https://github.com/airbrake/airbrake/pull/472#issuecomment-178527362.

PR #481, opened 2016-01-15
Add rubocop_todo.yml containing new offenses

kyrylo:
The Rubocop gem has been updated recently and we're now getting some
new offenses. In this commit we store them (and ignore, so the test
suite can pass) and then we can fix them later, one by one.


zefer:
LGTM

PR #487, opened 2016-01-18
rubocop: disable the Style/FrozenStringLiteralComment cop until Ruby 3

kyrylo:


zefer:
LGTM

(relates to airbrake/airbrake-ruby#34)

PR #489, opened 2016-01-19
Specify RAILS_ENV and RACK_ENV in capistrano rake task

thekendalmiller:
Restore some of the Capistrano functionality that is missing in the latest version. We might also want to add back `:airbrake_env`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Set RACK_ENV and RAILS_ENV in Capistrano 2 integration
+  ([#489](https://github.com/airbrake/airbrake/pull/489))
+
 ### [v5.0.3][v5.0.3] (January 19, 2015)
 
 * Improved RubyMine support
--- a/lib/airbrake/capistrano/tasks.rb
+++ b/lib/airbrake/capistrano/tasks.rb
@@ -39,6 +39,9 @@ else
               command = <<-CMD
                 cd #{config.release_path} && \
 
+                RACK_ENV=#{fetch(:rack_env, nil)} \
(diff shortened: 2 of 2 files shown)


kyrylo:
1. What do they do?
2. Why are they useful? You can already pass the env to ENVIRONMENT.

thekendalmiller:
I believe `ENVIRONMENT` is telling airbrake which environment to reset errors for. It's *not* telling what environment to run the rake task in.

`RAILS_ENV` tells what Rails environment that the rake task should be run in. In my case, the development servers are running in a 'qa' environment. When the Cap 2 airbrake task was running, it was running in 'production' and failing because there's not configurations in place for production on development servers.

I'm guessing `RACK_ENV` is the same idea, for rack apps(maybe sinatra?). I don't actually need it - I just restored it to be helpful.

kyrylo:
Thank you for the clarification. In this case you need to [prepend](https://github.com/airbrake/airbrake/blob/12aabd7c844e1ba2270a188efa36565b3f00519d/lib/airbrake/capistrano.rb#L81-L83) these variables to the `bundle exec rake airbrake:deploy` command. Currently you are passing them as "Rake env variables". It may be as is, but it's very confusing and I think it's better to implement it similar to the old approach (which is guaranteed to work).

I'd be happy to merge this with these changes.

thekendalmiller:
Thanks. I moved the variables.

kyrylo:
Thanks for the edit! I had to modify your commit a bit to make the test suite pass. Rubocop was complaining about high complexity of this method so I disabled the check.

I had to open another PR to be able to commit this. See https://github.com/airbrake/airbrake/pull/491

PR #495, opened 2016-02-02
rack: rely on airbrake-ruby's hostname information

kyrylo:
The new airbrake-ruby sends that information by default.


vmihailenco:
LGTM

PR #498, opened 2016-02-04
Check if user is truthy before trying to build from it

mangogogos:
Would otherwise cause an issue if the controller responds to current user but the value was e.g. nil


kyrylo:
Thanks for the PR. The build is red, though.

It would be nice to have a test for this change.

kyrylo:
Closed in favour of https://github.com/airbrake/airbrake/pull/498. Thanks!

zefer:
I think you meant #501 :smile:

kyrylo:
Oops, indeed, thanks! :)

PR #499, opened 2016-02-05
rack: attach more debugging information

kyrylo:


kyrylo:
PTAL

spec/integration/shared_examples/rack_examples.rb
```diff
@@ -101,6 +101,18 @@ def wait_for_a_request_with_body(body)
       it "features userAgent" do
         wait_for_a_request_with_body(/"context":{.*"userAgent":"Bot".*}/)
       end
+
+      it "features referer" do
+        wait_for_a_request_with_body(/"context":{.*"referer":"bingo.com".*}/)
+      end
+
+      it "contains HTTP headers" do
```
zefer: Should we also include a test for `CONTENT_TYPE`?

zefer:
LGTM

spec/integration/shared_examples/rack_examples.rb
kyrylo: This is an example for an error in someone else's app. We simply ensure here that we are able to report headers if there were any. We already check our own content type when we send a request to our server. We have that test in airbrake-ruby.

lib/airbrake/rack/notice_builder.rb
```diff
@@ -45,6 +54,9 @@ def add_context(notice)
 
         context[:url] = @request.url
         context[:userAgent] = @request.user_agent
+        context[:httpMethod] = @request.request_method
+        context[:referer] = @request.referer
+        context[:headers] = request_headers
```
vmihailenco: Airbrake API is free to throw away such properties, because they are not officially supported.

kyrylo: Let's support them.

PR #501, opened 2016-02-08
rack: check if user is truthy before trying to build from it

kyrylo:
Would otherwise cause an issue if the controller responds to current user but
the value was e.g. nil

Replaces #498 because the original author hasn't responded yet and I feel like it's a quite important addition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Airbrake Changelog
 
 ### master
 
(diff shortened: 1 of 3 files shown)


zefer:
:+1: LGTM, and thanks @mangogogos for reporting the issue!

mangogogos:
Thanks for getting this covered!

PR #503, opened 2016-02-09
rack: move the Rack env payload to 'environment'

kyrylo:
This is a follow-up to #499.

With the old approach the API can and probably will throw away our
data. Given that we're retrieving values about HTTP request from
the Rack env, let's attach it to the environment tab instead.


vmihailenco:
I still think that it should be in `Params` tab and have some prefix, but it is fine to fix when it becomes an issue :)

LGTM

kyrylo:
Thanks. I mostly prefer this solution because we have a tab that is not used at the moment.

PR #507, opened 2016-02-10
airbrake_env deployment notification option; fixes #506

### LINKED ISSUE #506: capistrano airbrake_env option no longer supported in Airbrake 5
Airbrake 5 has dropped the airbrake_env option for Capistrano deploys that was present in Airbrake 4.  This allowed the environment to be set differently than rails_env.  This is useful when running multiple "shards" of a production application that are used for different customers.  Shards have different versions deployed and are configured differently, so it is useful to separate their exceptions in Airbrake.  Deployments to one shard should not affect exceptions in another shard.

I think the only effect of this option is to control the argument to the rake task that notifies Airbrake of a deployment.  The Capistrano task used to look like:

    airbrake_env = fetch(:airbrake_env, fetch(:rails_env, fetch(:stage)))
    notify_command = "airbrake:deploy"
    notify_command << " TO=#{airbrake_env}"
    notify_command << " REVISION=#{fetch(:current_revision)} REPO=#{fetch(:repo_url)}"
    notify_command << " USER=#{local_user.strip.shellescape}"

which would let us define :airbrake_env in our deployment rb file.  Now in Airbrake 5 it looks like:

    airbrake:deploy USERNAME=#{Shellwords.shellescape(local_user)} \
                    ENVIRONMENT=#{fetch(:rails_env, fetch(:stage))} \
                    REVISION=#{fetch(:current_revision)} \
                    REPOSITORY=#{fetch(:repo_url)} \
                    VERSION=#{fetch(:app_version)}

and the airbrake_env option is no longer supported.  Please bring this back, as it's a useful setting.

wlipa:
Restores the airbrake_env Capistrano option from Airbrake 4, as discussed in #506.

### DIFF
--- a/lib/airbrake/capistrano/tasks.rb
+++ b/lib/airbrake/capistrano/tasks.rb
@@ -9,7 +9,7 @@ if defined?(Capistrano::VERSION) &&
           with rails_env: fetch(:rails_env, fetch(:stage)) do
             execute :rake, <<-CMD
               airbrake:deploy USERNAME=#{Shellwords.shellescape(local_user)} \
-                              ENVIRONMENT=#{fetch(:rails_env, fetch(:stage))} \
+                              ENVIRONMENT=#{fetch(:airbrake_env, fetch(:rails_env, fetch(:stage)))} \
                               REVISION=#{fetch(:current_revision)} \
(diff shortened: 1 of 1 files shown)


kyrylo:
:+1:

PR #510, opened 2016-02-12
Existing information in the notice's environment should be merged, not replaced

kimrgrey:
Information about ENV should be merged into the notice[:environment] to not replace already stored data. And `Airbrake::Rack::NoticeBuilder` should not replace lost information during building, too.

Please, take a look on this issue for additional details: https://git.io/vg1W0.

### DIFF
--- a/spec/unit/rack/notice_builder_spec.rb
+++ b/spec/unit/rack/notice_builder_spec.rb
@@ -31,5 +31,19 @@ RSpec.describe Airbrake::Rack::NoticeBuilder do
 
       expect(notice[:params]).to eq(params)
     end
+
+    it "adds CONTENT_TYPE, CONTENT_LENGTH and HTTP_* headers in the environment" do
+      headers = { "HTTP_HOST" => "example.com", "CONTENT_TYPE" => "text/html", "CONTENT_LENGTH" => 100500 }
+      notice_builder = described_class.new(headers.dup)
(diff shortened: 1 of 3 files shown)


kyrylo:
Could you also update the NoticeBuilder class, so it merges as well?

kimrgrey:
Sure. I'll add spec for this case as well to avoid possible regression in the future.

kimrgrey:
@kyrylo, done, take a look on this.

spec/unit/rack/notice_builder_spec.rb
```diff
@@ -31,5 +31,43 @@
...
+        "HTTP_HOST" => "example.com",
+        "CONTENT_TYPE" => "text/html",
+        "CONTENT_LENGTH" => 100500
+      }
+      notice_builder = described_class.new(headers.merge("X-SOME-HEADER" => "value"))
+      notice = notice_builder.build_notice(AirbrakeTestError.new)
+      expect(notice[:environment][:headers]).to eq(headers)
+    end
+
+    it "saves data that already has been added to the environment" do
```
kyrylo: Would the word `preserves` instead of `saves` describe this example better?

kimrgrey: @kyrylo, yep, changed and fixed up the commit.

kyrylo:
Looks fantastic, thanks :+1:

PR #514, opened 2016-02-19
Pass middleware class directly, not as a string

pschambacher:
There's some deprecation warnings on Rails 5.0.0.beta1 and .beta2 because of the middleware being passed as string. This should remove the deprecations.

@kyrylo maybe?

### DIFF
--- a/lib/airbrake/railtie.rb
+++ b/lib/airbrake/railtie.rb
@@ -14,16 +14,16 @@ module Airbrake
 
       middleware = if defined?(ActionDispatch::DebugExceptions)
         # Rails >= 3.2.0
-        "ActionDispatch::DebugExceptions"
+        ActionDispatch::DebugExceptions
       else
         # Rails < 3.2.0
-        "ActionDispatch::ShowExceptions"
+        ActionDispatch::ShowExceptions
       end
 
       app.config.middleware.insert_after middleware,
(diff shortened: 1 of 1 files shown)


kyrylo:
I'm not sure about this. If you're going to upgrade to Rails 5, why not to upgrade Airbrake as well? :P

pschambacher:
We have a lot of projects using Airbrake and considering the effort of migrating the gem, we're delaying it and we'd rather keep using the v4 for a while since it's still supported for now :smile:

kyrylo:
The code presented this PR will probably break older Rails apps since from what I understand they want a string. I am still not sure about this change. Rails 5 hasn't been officially released, so you still have time to upgrade (it's not a lot of work and we have a [migration guide](https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md)).

kyrylo:
I'd like to clarify that although it's easy to fix this PR and merge, the problem is that when you stumble upon a bug in Rails 5 & Airbrake 4, then we would need to fix it. It's already hard to maintain two branches and I consider this PR an upgrade.

As stated in the readme, the `airbrake-v4` branch is feature frozen, sorry.
I highly recommend upgrading to Airbrake 5 (it's also much more better than Airbrake 4).

PR #515, opened 2016-02-24
rack: add support for multiple notifiers

kyrylo:
Work in progress.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Added support for multiple notifiers for Rack middleware
+  ([#515](https://github.com/airbrake/airbrake/pull/515))
+
 ### [v5.5.0][v5.5.0] (September 14, 2016)
 
 * Started depending on
--- a/README.md
+++ b/README.md
@@ -212,6 +212,10 @@ project's dashboard for a new error.
 curl localhost:9292
 ```
 
+If your Sinatra app consists of subprojects and you want to capture errors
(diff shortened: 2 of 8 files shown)


kyrylo:
We couldn't figure out best way to implement this and [agreed to close this for now](https://github.com/airbrake/airbrake-ruby/pull/62#issuecomment-206788580).

jonahx:
@kyrylo I may be missing something obvious, since I only just glanced at the source here:

https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rack/middleware.rb

but isn't it just a matter of:

1. passing the name of the config to the Middleware initializer, and saving that, say as `@config_name`.
2. changing line 24 to `notify_airbrake(ex, env, @config_name)`

Then in your sinatra app: `use Airbrake::Rack::Middleware, :some_config`

kyrylo:
Your described scenario is achieved with this PR, but with a slightly different API. I like what you suggested more because `use Airbrake::Rack::Middleware, :some_config` seems to be simpler. It's been a while and I need to refresh my memory, so I am going to investigate this feature again and report back shortly.

kyrylo:
@jonahx the support has been added. Could you please test this PR with your project and tell me if it meets your expectations?

jonahx:
@kyrylo Thanks, this looks great!  Exactly what I wanted.

AwolDes:
Closing the loop on this. Named notifiers were removed at some point: https://github.com/airbrake/airbrake-ruby/pull/493

PR #516, opened 2016-02-24
Added method add_metadata to Airbrake::Rack::NoticeBuilder.

kimrgrey:
This method does nothing by default, but it can be overridden to add any custom data from RACK environment to the notice.


kyrylo:
Thanks! :hand: 

Exactly what I envisioned. However, the lack of 1.9 support is indeed something that I forgot about. I just also realised this solution is not going to work if you define multiple hooks (imagine you depended on something that also defines this hook).

Let's focus on https://github.com/airbrake/airbrake/pull/517.

Issue #509, opened 2016-02-12
Impossible to add addtional info about RACK env to the notice


kimrgrey:
As I understand from code, there is only one point where information from RACK environment is available for `Notice`. It is `Rack::NoticeBuilder`. But it adds only specific type of information that is predefined. So, it is impossible to add `REMOTE_IP` or something like this to the created notice. Am I missed something? May be it's good idea to allow users extend `NoticeBuilder`?

kyrylo:
That's correct. The idea is that the defaults should be good enough for everyone. I will think about the way to allow this, but if you have any ideas or examples, feel free to share. It's not straightforward.

kimrgrey:
Well, it's almost not possible to satisfy everyone :smile: May be we it's good idea to allow users to create their own builders that is inherited from `Rack::NoticeBuilder` and after it configure `Airbrake` to use it?

kimrgrey:
Hi, @kyrylo! What about this stuff? We really need to add some additional headers in our project. And I don't want to use monkey patching for it :disappointed:

kyrylo:
Hey, @kimrgrey! Don't worry, you are not forgotten. I need some more time. This will be resolved during this week.

kimrgrey:
Good news, thanks! I will follow up :wink:

kyrylo:
What do you think about the following API:

```ruby
# config/initializers/airbrake.rb
# Standard config... and then...

module AirbrakeNoticeBuilderMetadata
  def add_metadata(notice)
    notice[:params][:remoteIp] = @rack_env['REMOTE_IP']
  end
end

class Airbrake::Rack::NoticeBuilder
  prepend AirbrakeNoticeBuilderMetadata
end
```

Plain Ruby, no shenanigans. You get access to everything you wish for. Might be a bit bare bones.

kimrgrey:
Well, it looks like officially allowed monkey patching =) But, as for me, it's really simple solution, it's easy to add to the  current code and it solves the problem.

kimrgrey:
@kyrylo, should I send a PR with this change?

kimrgrey:
- [ ] account_accesses_controller.rb
- [ ] affiliates_controller.rb
- [ ] agreement_reports_controller.rb
- [ ] agreements_controller.rb
- [ ] announcements/announcements_controller.rb
- [ ] announcements/categories_controller.rb
- [ ] api_controller.rb
- [ ] application_controller.rb
[...]

PR #517, opened 2016-02-24
Introduced user defined builders in Airbrake::Rack::NoticeBuilder (#509)

kimrgrey:
Allows users to add their own builders for notice. So, for example, if it's needed to add a `REMOTE_IP` in the notice's params it can be easily done in standard initializer for airbrake like this:

```ruby
Airbrake.add_rack_builder |notice, request|
  notice[:params][:remoteIp] = request.env['REMOTE_IP']
end
```

### DIFF
--- a/lib/airbrake.rb
+++ b/lib/airbrake.rb
@@ -19,3 +19,25 @@ require 'airbrake/rake/task_ext' if defined?(Rake::Task)
 require 'airbrake/resque/failure' if defined?(Resque)
 require 'airbrake/sidekiq/error_handler' if defined?(Sidekiq)
 require 'airbrake/delayed_job/plugin' if defined?(Delayed)
+
+##
+# This module extends original module Airbrake from airbrake-ruby and
+# serves as a namespace for other classes and modules.
+module Airbrake
+  class << self
+    ##
+    # Allows users to add their own notice's builders. Useful if it's needed to
+    # attach some info from the rack environment or request to the notice.
+    #
+    # @example Adds remote ip from the rack_env to the notice
+    #   Airbrake.add_rack_builder |notice, request| do
+    #     notice[:params][:remoteIp] = request.env['REMOTE_IP']
+    #   end
+    #
+    # @yieldparam notice [Airbrake::Notice] notice that will be sent to the Airbrake
+    # @yieldparam request [Rack::Request] current rack request
+    def add_rack_builder(&block)
+      Airbrake::Rack::NoticeBuilder.add_builder(&block)
+    end
+  end
+end
--- a/lib/airbrake/rack/notice_builder.rb
+++ b/lib/airbrake/rack/notice_builder.rb
@@ -4,6 +4,20 @@ module Airbrake
     # A helper class for filling notices with all sorts of useful information
     # coming from the Rack environment.
     class NoticeBuilder
+      @builders = []
+
+      class << self
+        ##
+        # @return [Array<#call>] the list of notice's builders
+        attr_reader :builders
+
+        ##
+        # Adds user defined builders to the chain.
+        def add_builder(&block)
+          @builders << block
+        end
+      end
+
       ##
       # @return [Array<String>] the prefixes of the majority of HTTP headers in
       #   Rack (some prefixes match the header names for simplicity)
@@ -16,20 +30,7 @@ module Airbrake
       ##
       # @param [Hash{String=>Object}] rack_env The Rack environment
       def initialize(rack_env)
-        @rack_env = rack_env
         @request = ::Rack::Request.new(rack_env)
-        @controller = rack_env['action_controller.instance']
-        @session = @request.session
-        @user = Airbrake::Rack::User.extract(rack_env)
-
-        @framework_version =
-          if defined?(::Rails)
-            "Rails/#{::Rails.version}"
-          elsif defined?(::Sinatra)
-            "Sinatra/#{Sinatra::VERSION}"
-          else
-            "Rack.version/#{::Rack.version} Rack.release/#{::Rack.release}"
-          end.freeze
       end
 
       ##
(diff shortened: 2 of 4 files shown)


kyrylo:
I quite like this, but I am not a fan of `instance_exec`. If you're going to read the code for the first time, you will definitely question `@rack_env`. It's not obvious where it comes from.

kyrylo:
I am not sure about the name, `filters`. Maybe it makes sense to rename this to `builders` (`add_builder`)?

kyrylo:
How about this API, so it doesn't look like we're messing with library's internals?

```ruby
Airbrake.add_rack_builder do |notice, rack_env|
  # ...
end
```

kimrgrey:
> How about this API, so it doesn't look like we're messing with library's internals?

Yeah, it looks better. I was thinking about it. But in this case if we need, for example, `@request`  we should create it from `rack_env` in every builder separately. Is it fine?

kyrylo:
Could we yield it instead, if it's common?

kimrgrey:
> Could we yield it instead, if it's common?

Yep, it's better. I've implemented each builder as a block that takes two arguments - notice and request. So, we have an access to the `rack_env` through `request.env`. But I'm not sure where is a good place to define `Airbrake.add_rack_builder`? Should I define it in `airbrake.rb` or somewhere else?

kyrylo:
Wise decision!

`airbrake.rb` sounds good to me.

kimrgrey:
@kyrylo, I've implemented `add_rack_builder` and  added some documentation. It looks finished to me. WDYT?

lib/airbrake.rb
```diff
@@ -19,3 +19,28 @@
 require 'airbrake/resque/failure' if defined?(Resque)
 require 'airbrake/sidekiq/error_handler' if defined?(Sidekiq)
 require 'airbrake/delayed_job/plugin' if defined?(Delayed)
+
+##
+# This module defines the Airbrake API and serves as a namespace for other
+# classes and modules.
+##
```
kyrylo: Nitpick: we don't end comments with `##`, so I suggest to remove them (but keep `##` in the beginnings), so it looks consistent.

lib/airbrake.rb
```diff
@@ -19,3 +19,28 @@
...
+    # add a REMOTE_IP to the notice's params it can be easily done in standard
+    # initializer for Airbrake.
+    #
+    # @example Ignore all notices
+    #   Airbrake.add_rack_builder |notice, request|
+    #     notice[:params][:remoteIp] = request.env['REMOTE_IP']
+    #   end
+    #
+    # @yield [notice, request] The builder block
+    # @note Once a builder was added, there's no way to delete it
```
kyrylo: I would remove this because you surely can delete them with `Airbrake::Rack::NoticeBuilder.builders.clear` :)

lib/airbrake.rb
```diff
@@ -19,3 +19,28 @@
...
+##
+module Airbrake
+  class << self
+    ##
+    # Allows users to add their own notice's builders. For example, if it's needed to
+    # add a REMOTE_IP to the notice's params it can be easily done in standard
+    # initializer for Airbrake.
+    #
+    # @example Ignore all notices
+    #   Airbrake.add_rack_builder |notice, request|
```
kyrylo: Missing `do`.

lib/airbrake.rb
```diff
@@ -19,3 +19,28 @@
...
+# classes and modules.
+##
+module Airbrake
+  class << self
+    ##
+    # Allows users to add their own notice's builders. For example, if it's needed to
+    # add a REMOTE_IP to the notice's params it can be easily done in standard
+    # initializer for Airbrake.
+    #
+    # @example Ignore all notices
```
kyrylo: Misleading title.

lib/airbrake.rb
```diff
@@ -19,3 +19,28 @@
 require 'airbrake/resque/failure' if defined?(Resque)
 require 'airbrake/sidekiq/error_handler' if defined?(Sidekiq)
 require 'airbrake/delayed_job/plugin' if defined?(Delayed)
+
+##
+# This module defines the Airbrake API and serves as a namespace for other
```
kyrylo: Could you please mention that it extends the original class from `airbrake-ruby`? That is, we reopen its definition here.

lib/airbrake.rb
```diff
@@ -19,3 +19,28 @@
...
+# classes and modules.
+##
+module Airbrake
+  class << self
+    ##
+    # Allows users to add their own notice's builders. For example, if it's needed to
```
kyrylo: Could you paraphrase this paragraph so it looks more general? Just mention that with help of this you can attach additional information from the rack environment/request.

lib/airbrake.rb
```diff
@@ -19,3 +19,28 @@
...
+    # add a REMOTE_IP to the notice's params it can be easily done in standard
+    # initializer for Airbrake.
+    #
+    # @example Ignore all notices
+    #   Airbrake.add_rack_builder |notice, request|
+    #     notice[:params][:remoteIp] = request.env['REMOTE_IP']
+    #   end
+    #
+    # @yield [notice, request] The builder block
```
kyrylo: Could you specify [`@yieldparam`](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#yieldparam) instead?

lib/airbrake/rack/notice_builder.rb
```diff
@@ -39,64 +43,67 @@ def initialize(rack_env)
       # @return [Airbrake::Notice] the notice with extra information
       def build_notice(exception)
         notice = Airbrake.build_notice(exception)
-
-        add_context(notice)
-        add_session(notice)
-        add_params(notice)
-        add_environment(notice)
-
+        NoticeBuilder.builders.each { |filter| filter.call(notice, @request) }
```
kyrylo: `s/filter/builder/`

spec/airbrake_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+RSpec.describe Airbrake do
+  describe "#add_rack_builder" do
```
kyrylo: `s/#add_rack_builder/.add_rack_builder/`

spec/airbrake_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+    it "adds new builder to the chain" do
+      builder = proc { |_, _| nil }
+      expect(Airbrake::Rack::NoticeBuilder).to receive(:add_builder) do |_, &block|
```
kyrylo: Could you verify that the builder counter changes by 1 instead (or as well)?

kimrgrey: Well, I think it's not good idea to really change the builder's list here because in this case it will impact all other specs.

lib/airbrake/rack/notice_builder.rb
```diff
@@ -4,6 +4,23 @@ module Rack
     # A helper class for filling notices with all sorts of useful information
     # coming from the Rack environment.
     class NoticeBuilder
+      @builders = []
+
+      class << self
+        ##
+        # @return [Array<Callable>] the list of notice's builders
```
kyrylo: I think it should be `[Array<#call>]` instead.

spec/unit/rack/notice_builder_spec.rb
```diff
@@ -69,5 +69,15 @@
       notice = notice_builder.build_notice(AirbrakeTestError.new)
       expect(notice[:environment]["SOME_KEY"]).to eq("SOME_VALUE")
     end
+
+    it "applies custom builders for notice" do
```
kyrylo: `"runs builders against notices"` sounds more understandable IMO (or something similar). "Applies" is a bit confusing.

lib/airbrake/rack/notice_builder.rb
```diff
@@ -39,64 +43,67 @@ def initialize(rack_env)
...
-        add_environment(notice)
-
+        NoticeBuilder.builders.each { |filter| filter.call(notice, @request) }
         notice
       end
 
-      private
-
-      def add_context(notice)
+      add_builder do |notice, request|
```
kyrylo: Could you add small comments explaining what a builder adds? So people don't have to read code to understand what it attaches?

For example:

```
# Adds context.
add_builder do |notice, request|
  ...
```

spec/airbrake_spec.rb
kyrylo: Sure, understandable, that said, you could cleanup after you've done checking. It's important to test this properly because it's public API.

kyrylo:
Looks pretty cool! Just needs a couple minor tweaks and it's ready :)

kimrgrey:
Good news! Thanks for the review! :+1: I'll finish everything tomorrow.

kimrgrey:
@kyrylo, I've fixed all your notices. Please, check it and I will squash commits.

lib/airbrake.rb
```diff
@@ -19,3 +19,26 @@
...
+    ##
+    # Allows users to add their own notice's builders. Useful if it's needed to
+    # attach some info from the rack environment or request to the notice.
+    #
+    # @example Adds remote ip from the rack_env to the notice
+    #   Airbrake.add_rack_builder |notice, request| do
+    #     notice[:params][:remoteIp] = request.env['REMOTE_IP']
+    #   end
+    #
+    # @yield block that will be used to build a notice
```
kyrylo: Please remove this line (not needed if you specify `@yieldparam`).

lib/airbrake/rack/notice_builder.rb
```diff
@@ -4,6 +4,22 @@ module Rack
...
+
+      class << self
+        ##
+        # @return [Array<#call>] the list of notice's builders
+        attr_reader :builders
+
+        ##
+        # Adds user defined builders to the chain.
+        #
+        # @yield [notice, request] The builder block
```
kyrylo: :)

Probably just remove this line as it was already documented.

kyrylo:
Looks even better. Just needs two super minor documentation fixes and it's ready!

kimrgrey:
Yep =) Removed outdated comments for `yield`, squashed commits.

kyrylo:
:+1:

PR #518, opened 2016-02-27
Document Airbrake.add_rack_builder

kyrylo:


mmcdaris:
Nice one, Now the docs have caught up to the comments in [lib/airbrake.rbL32-35](https://github.com/airbrake/airbrake/blob/master/lib/airbrake.rb#L32-L35) LGTM!

kyrylo:
:+1:

PR #519, opened 2016-02-27
Fix Airbrake::Rack::Middleware regression

ajvondrak:
In v4, Airbrake::Rack would check for framework-saved exceptions in the
Rack env under rack.exception or sinatra.error:
https://github.com/airbrake/airbrake/blob/34fcbe8e6aae20aad0e4f5174df16945dd6f1356/lib/airbrake/rack.rb#L60

Now v5's Airbrake::Rack::Middleware looks under
action_dispatch.exception or sinatra.error. I'm not sure whether Rails
or Sinatra ever set env['rack.exception'] on their own. That key does
not appear to be part of the Rack standard, and I can't find any trace
of the string 'rack.exception' in the Rails codebase, or even in a `git
log -G 'rack\.exception'`.

Presumably Rails (or maybe a middleware it depends on?) set that key at
some point, since Airbrake v4 checked for it. I also presume v5
continues to work fine with Rails and Sinatra. However, the
rack.exception key still exists out in the ecosystem. In particular,
upgrading to Airbrake v5 introduced broke my Goliath app:
- https://github.com/postrank-labs/goliath/blob/4d7bd448bf292802d30d971be97f49d191c3359f/lib/goliath/constants.rb#L25
- https://github.com/postrank-labs/goliath/blob/4d7bd448bf292802d30d971be97f49d191c3359f/lib/goliath/api.rb#L180

This patch reintroduces the check for rack.exception. This may be
hinting that we need an overall cleaner way of handling frameworks'
weird ways of stashing middleware exceptions, but this was the simplest
path towards getting my app working again.

### DIFF
--- a/lib/airbrake/rack/middleware.rb
+++ b/lib/airbrake/rack/middleware.rb
@@ -26,9 +26,7 @@ module Airbrake
         end
         # rubocop:enable Lint/RescueException
 
-        # The internal framework middlewares store exceptions inside the Rack
-        # env. See: https://goo.gl/Kd694n
-        exception = env['action_dispatch.exception'] || env['sinatra.error']
+        exception = framework_exception(env)
         notify_airbrake(exception, env) if exception
 
         response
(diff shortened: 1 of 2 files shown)


lib/airbrake/rack/middleware.rb
```diff
@@ -40,6 +38,14 @@ def notify_airbrake(exception, env)
         notice = NoticeBuilder.new(env).build_notice(exception)
         Airbrake.notify(notice)
       end
+
+      # The internal framework middlewares store exceptions inside the Rack
+      # env. See: https://goo.gl/Kd694n
+      def framework_exception(env)
+        env['rack.exception'] ||
```
kyrylo: Could you please move this to the end of the `||` pipeline, so it has least priority. As of now it seems like it's the default key that comes from Rack, which is not true. Could you please also explicitly mention that it's a Goliath thing (and link to it, too)?

kyrylo:
Hey! Thanks for the PR. It was made so intentionally.

While writing this code I couldn't find any trails of `'rack.exception'` as well, so that's why I deleted it. Now it's evident where it comes from, so it deserves to be included.

I am happy to merge this with a couple fixes that I mentioned above.

ajvondrak:
Thanks for the quick (same day!) turnaround time. Commit amended to address your comments. Let me know how it looks.

kyrylo:
Looks great, thanks!

PR #533, opened 2016-03-23
rails/active_job: amend `component` and `action` to fix grouping

### LINKED ISSUE #532: Same exceptions occurred from ActiveJob have different action values
Hi, I'm using sidekiq as queue adapter of active_job.
Recently I noticed every occurrence of exceptions like `ActiveRecord::RecordInvalid` is grouped separately on airbrake dashboard.
The reason turned out to be different action value of each occurrence differs since exception handler of activejob is assigning `job_id` for action value.(https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rails/active_job.rb#L13)
I think name of the worker class should be an action value instead of job_id since it's uuid as in sidekiq middleware(https://github.com/airbrake/airbrake/blob/master/lib/airbrake/sidekiq/error_handler.rb).

kyrylo:
Fixes #532 (Same exceptions occurred from ActiveJob have different
action values)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Fixed grouping for the ActiveJob integration
+  ([#533](https://github.com/airbrake/airbrake/pull/533))
+
 ### [v5.2.1][v5.2.1] (March 21, 2016)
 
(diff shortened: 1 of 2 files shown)


kyrylo:
PTAL

zefer:
LGTM

PR #534, opened 2016-03-23
Appraisals: fix build for Ruby 1.9.3

kyrylo:
Newest `mime-types` gem dropped support for 1.9.3, so our build has
started failing. Locking down the gem version helps to fix it.


kyrylo:
Merging as a quick fix.

PR #536, opened 2016-03-28
rake: handle response errors

kyrylo:
If a project was configured with bad project ID or project key, the API
would return an error response. The Rake task, however, would print the
success text.

Instead, handle API errors and print a more relevant message. Example:

```
% bundle exec rake airbrake:test
Error occurred: Project not found or access denied.
Make sure that your Project ID and Project Key are correct:
https://github.com/airbrake/airbrake-ruby#project_id--project_key
```

### DIFF
--- a/lib/airbrake/rake/tasks.rb
+++ b/lib/airbrake/rake/tasks.rb
@@ -15,7 +15,23 @@ namespace :airbrake do
       "#{name}:\n  " + [cfg, filters].pretty_inspect
     end.join("\n")
 
-    if response
+    if !response
+      puts <<NORESPONSE
+Couldn't send a test exception. There are two reasons for this:
+
+1. Airbrake ignored this exception due to misconfigured filters
+2. Airbrake was configured to ignore the '#{Rails.env}' environment.
+   To fix this try one of the following:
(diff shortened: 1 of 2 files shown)


lib/airbrake/rake/tasks.rb
```diff
@@ -15,7 +15,23 @@
       "#{name}:\n  " + [cfg, filters].pretty_inspect
     end.join("\n")
 
-    if response
+    if !response
+      puts <<NORESPONSE
```
shifi: What do you think of [adding some margin based on which to trim the excessive whitespace](https://github.com/bbatsov/ruby-style-guide#heredocs)?

```ruby
puts <<-NORESPONSE.gsub(/^\s+\|/, '')
  |Couldn't send a test exception. There are two reasons for this:
  |1. Airbrake ignored this exception due to misconfigured filters
  |...
NORESPONSE

shifi:
LGTM

lib/airbrake/rake/tasks.rb
kyrylo: Sounds reasonable. Fixed.

PR #537, opened 2016-03-29
Migration_Guide: mention deprecation of 'filter_parameters'

kyrylo:
A customer was confused that we didn't mention this in the guide.


shifi:
LGTM

PR #538, opened 2016-03-30
rake: attach env information to the Heroku deploy hook

kyrylo:
At the moment errors cannot be resolved automatically on deploy because
Heroku doesn't send the environment information. To work around this we
attach environment to the URL

### DIFF
--- a/lib/airbrake/rake/tasks.rb
+++ b/lib/airbrake/rake/tasks.rb
@@ -100,7 +100,14 @@ namespace :airbrake do
 
     exit!(1) if [id, key].any?(&:nil?)
 
+    unless (env = heroku_env['RAILS_ENV'])
+      env = 'production'
+      puts "Airbrake couldn't identify your app's environment, so the '#{env}'" \
+           " environment will be used."
+    end
+
     url = "https://airbrake.io/api/v3/projects/#{id}/heroku-deploys?key=#{key}"
+    url << "&environment=#{env}"
 
     command = %(heroku addons:create deployhooks:http --url="#{url}")
     command << " --app #{app}" if app


lib/airbrake/rake/tasks.rb
```diff
@@ -100,7 +100,10 @@
 
     exit!(1) if [id, key].any?(&:nil?)
 
+    env = heroku_env['RAILS_ENV'] || ENV['RAILS_ENV']
```
zefer: I think the fallback `ENV['RAILS_ENV']` in most cases will be `development`, since that's where this rake task will normally be run from. Maybe simply `heroku` is a less confusing default, but I may be missing something?

kyrylo: Simply `heroku` won't cut it because we need to be able to resolve errors automatically, so the value should match with what users have. How about:

```ruby
env = heroku_env['RAILS_ENV'] || 'production'
```

zefer: Sounds better. If it causes confusion, we could output a warning message when we don't know the heroku env, explaining we picked a default and why the value should match. This would allow them to edit the deploy hook if the default was inappropriate.

Just curious, does a dynamic URL value for the deploy hook? E.G "some-url?#{ENV['RAILS_ENV']}"?

kyrylo: Applied this change and added a warning.

> Just curious, does a dynamic URL value for the deploy hook? E.G "some-url?#{ENV['RAILS_ENV']}"?

No, it shouldn't work because what comes after the pound sign is not going to be sent to the server.

kyrylo: PTAL

zefer:
LGTM

PR #542, opened 2016-04-14
rails/active_job: be synchronous is the adapter is resque

### LINKED ISSUE #540: Configuration for ActiveJob + Resque
* Airbrake version: 5.2.3
* Airbrake-ruby version: 1.2.2
* Ruby version: 2.2.4
* Framework name & version: Rails 4.2.6
* ActiveJob version: 4.2.6
* Resque version: 1.26.0

#### Airbrake config

```ruby
Airbrake.configure do |c|
  c.project_id = ENVVARS["AIRBRAKE_PROJECT_ID"]
  c.project_key = ENVVARS["AIRBRAKE_PROJECT_KEY"]

  c.root_directory = Rails.root
  c.logger = Rails.logger
  c.environment = Rails.env
  c.ignore_environments = %w(development test)
end

  unless ENVVARS["AIRBRAKE_REPORTING"] == "enabled"
    notice.ignore!
  end
end
```

#### Description

We're using Airbrake with Resque run via ActiveJob. In the Airbrake README, it says that no configuration should be required when using ActiveJob.

We weren't getting error reports from background jobs. We looked into it and figured out that Resque is forking its worker process for each job that it runs. With Airbrake sending notifications asynchronously, the forked process was exiting before the Airbrake thread could send the notification.

We fixed this by adding this to our config:

```ruby
Resque.after_fork do
  at_exit { Airbrake.close }
end
```

We also had to set `ENV["RUN_AT_EXIT_HOOKS"]` for Resque.

It took a while to figure this out, so I'd like to contribute something to Airbrake to save other people who run into this same issue some time. Should there be an update to the README for this? I couldn't think of a way to update the code, but I'm happy to create a PR if somebody has a suggestion.

kyrylo:
Fixes #540 (Configuration for ActiveJob + Rescue).
 
This goes along with our Resque integration, which became synchronous.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Fixed bug in the ActiveJob+Resque integration, where the gem couldn't report
+  any exceptions ([#542](https://github.com/airbrake/airbrake/pull/542))
+
 ### [v5.2.3][v5.2.3] (April 5, 2016)
 
 * Fixed bug in the Rake integration where it couldn't display error message
--- a/lib/airbrake/rails/active_job.rb
+++ b/lib/airbrake/rails/active_job.rb
@@ -6,6 +6,11 @@ module Airbrake
       extend ActiveSupport::Concern
 
       included do
+        if ::Rails.application.config.respond_to?(:active_job)
+          active_job_cfg = ::Rails.application.config.active_job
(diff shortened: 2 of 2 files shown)


kyrylo:
PTAL

kyrylo:
The user confirmed this worked for them.

shifi:
LGTM

Issue #546, opened 2016-05-04
Disable Airbrake when not configured?


khustochka:
In Airbrake v4 I used to do this:

````ruby
# config/initializers/airbrake.rb
if ENV['AIRBRAKE_API_KEY']
  Airbrake.configure do |config|
    config.api_key = ENV['AIRBRAKE_API_KEY']
  end
end
````

And it worked as I expected: if AIRBRAKE_API_KEY environment variable was not set Airbrake was silent. It was OK in development, test, and on staging servers that ran in production mode but had no need in error reporting.

After upgrade to v5 I wanted to do the same:

````ruby
# config/initializers/airbrake.rb
if ENV['AIRBRAKE_PROJECT_ID'] && ENV['AIRBRAKE_API_KEY']
  Airbrake.configure do |config|
    config.project_id = ENV['AIRBRAKE_PROJECT_ID']
    config.project_key = ENV['AIRBRAKE_API_KEY']
    config.environment = Rails.env
  end
end
````

But this time it did not work, every exception was replaced with the notorious `Airbrake::Error (the 'default' notifier isn't configured)`! So now if I include airbrake in the project I am forced to always configure it. Even if I want to make use of `ignore_environments` I have to provide project id and key.

As a workaround I tried the following:
1. Declare `gem "airbrake", require: false` in `Gemfile`
2. In `config/initializers/airbrake.rb`:

````ruby
if ENV['AIRBRAKE_API_KEY'] && ENV['AIRBRAKE_API_KEY']
  require "airbrake"
  Airbrake.configure do |config|
    ......
  end
end
````
But this did not work with an opposite effect: errors were not sent to airbrake even when keys were properly provided. I may spent some time later to investigate why exactly this happens.

So... I do not know if this was intended behaviour or not, but for me it seems reasonable to have a way to completely disable or silence Airbrake even if the gem is included in the project. Personally I would prefer the way it worked in v4, if you do not call Airbrake.configure it is not activated, but there may be other possible solutions.

**Comment by kyrylo** [comment]
Hey, @khustochka! Would the following code be enough to meet your expectations?

```
Airbrake.add_filter(&:ignore!) unless ENV['AIRBRAKE_API_KEY']
```

khustochka:
Thanks, I was able to make it into a workaround. 

As I understand to call `Airbrake.add_filter` notifier should already be configured, and when I am configuring it I cannot leave project key and id blank, so I have to replace it with dummy values:

````
[...]

PR #547, opened 2016-05-09
Adjust integrations to return early when 'build_notice' returns 'nil'

kyrylo:
This change along with https://github.com/airbrake/airbrake-ruby/pull/75
fix https://github.com/airbrake/airbrake/issues/546.

Due to the breaking change in airbrake-ruby we need to update some of
our integrations that use `Airbrake.build_notice`.

/cc @khustochka (could you please verify if `airbrake-ruby` master with this PR work for you? I've done some manual testing and it is works good for me).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -71,3 +71,6 @@ Performance/StartWith:
 
 Style/Alias:
   Enabled: true
+
+Style/MultilineArrayBraceLayout:
+  Enabled: true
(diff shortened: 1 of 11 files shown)


.rubocop.yml
```diff
@@ -71,3 +71,6 @@ Performance/StartWith:
 
 Style/Alias:
   Enabled: true
+
+Style/MultilineArrayBraceLayout:
+  Enabled: true
```
kyrylo: Rubocop started failing out of the blue, so I had to fix this as well: https://circleci.com/gh/airbrake/airbrake/688

kyrylo:
PTAL

khustochka:
This works for me as expected. Thank you.

vmihailenco:
LGTM

PR #548, opened 2016-05-10
gemspec: update airbrake-ruby dependency to 1.3.0

kyrylo:

PR #550, opened 2016-05-10
Release v5.3.0

kyrylo:
Why v5.3.0 and not v5.2.4? Well, it's because airbrake-ruby's MINOR
version was bumped. SemVer doesn't define what to do in this situation.
This gem's API stays the same. However the logic here is that users are
directly affected by this change because airbrake-ruby doesn't raise
an important exception that it used to raise.

Depends on https://github.com/airbrake/airbrake/pull/548 & https://github.com/airbrake/airbrake/pull/549.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Airbrake Changelog
 
 ### master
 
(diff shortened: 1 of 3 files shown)


shifi:
LGTM

PR #559, opened 2016-06-14
Fix params cleaner for deep nested arrays

flexoid:
Issue Type: _Bug Report_
Airbrake Gem Version: **4.3.6**

Fix `#filter` method in `ParamsCleaner` class to be able to process nested arrays with any nesting level.
This error occurs in our sidekiq worker with unknown exact steps, but it can be easily reproduced in tests (which is done in this pull request).

According to [this page](https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md), v4 is still supported and, I hope, is open for bug fixes, as we would like to avoid migration to v5 for our current production apps.

Thanks.

### DIFF
--- a/lib/airbrake/utils/params_cleaner.rb
+++ b/lib/airbrake/utils/params_cleaner.rb
@@ -97,16 +97,15 @@ module Airbrake
         end
 
         def filter(hash)
+          return hash unless hash.is_a?(Hash)
+
           hash.each do |key, value|
-            if hash.is_a?(Hash) && filter_key?(key)
+            if filter_key?(key)
               hash[key] = "[FILTERED]"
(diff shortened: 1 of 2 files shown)


kyrylo:
Yeah, we do support v4 still. Thanks for the fix. The code has also became a bit simpler. 👍

flexoid:
Thanks for the quick merge. Can I ask if you are going to perform release in the near future?

kyrylo:
I plan to release it today. I'll ping you here.

kyrylo:
Just released `v4.3.7`. Enjoy! https://rubygems.org/gems/airbrake/versions/4.3.7

flexoid:
Cool, thanks!

PR #562, opened 2016-06-20
Use `clear_active_connections!` instead of `release_connection`

yui-knk:
`ActiveRecord::Base.connection_pool.release_connection` handles
only one of connection_pools which belongs to `ActiveRecord::Base`.
When developers use `establish_connection` and `User` model uses
an established connection, we should use `clear_active_connections!`
to release all connection_pools.

### DIFF
--- a/lib/airbrake/rails/controller_methods.rb
+++ b/lib/airbrake/rails/controller_methods.rb
@@ -137,8 +137,8 @@ module Airbrake
       ensure
         # The Airbrake middleware is first in the chain, before ActiveRecord::ConnectionAdapters::ConnectionManagement
         # kicks in to do its thing. This can cause the connection pool to run out of connections.
-        if defined?(ActiveRecord::Base) && ActiveRecord::Base.respond_to?(:connection_pool)
-          ActiveRecord::Base.connection_pool.release_connection
+        if defined?(ActiveRecord::Base) && ActiveRecord::Base.respond_to?(:clear_active_connections!)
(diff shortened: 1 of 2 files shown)


yui-knk:
For example `ConnectionManagement ` of Active Record uses `clear_active_connections!`

https://github.com/rails/rails/blob/v4.2.6/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L645

kyrylo:
Hey, @yui-knk! It's nice to see you here :)
Just curious, did you hit any bugs with the current code?
We will probably also need to port this method over to airbrake-v5 as well.

kyrylo:
A friendly ping.

yui-knk:
I am very happy to meet you :)
Sorry for my late response.

> did you hit any bugs with the current code?

Sure I hit bugs related to leaking of connection_pools, but I found out airbrake is not a cause.
However I still feel `clear_active_connections!` is better than `release_connection `, because `clear_active_connections!` release all connection_pools :)

kyrylo:
I just looked at the `clear_active_connections!` implementation, and it makes sense. Thanks! 👍

yui-knk:
😄

kyrylo:
Just released your fix: https://rubygems.org/gems/airbrake/versions/4.3.8
I recommend to update to v5, though. The fix for the same possible issue will be released today, too :)

yui-knk:
Thanks for your quick action!😺

PR #566, opened 2016-06-24
README: remove issue stats badges

kyrylo:
I really like the idea of them, but unfortunately they are very
unreliable. They often fail to load because the main app stops working.

airbrake airbrake the official airbrake library for ruby applications 2016-06-24 18-21-41


vmihailenco:
:+1:

PR #567, opened 2016-06-24
rack: release active connection

kyrylo:
Similar to #562. We have to have this extra safety measure to ensure
we don't leak DB connections. More information:
https://blog.codefront.net/2009/06/15/activerecord-rails-metal-too-many-connections/

In normal Rails code we don't need to do that because Rails has a
[middleware][1] to deal with this issue. Our Rack middleware kicks in
earlier, so we have to deal with this on our own.

[1]: https://github.com/rails/rails/blob/107f4282bbfabc011d5ad3bcf3fb3c6fb812ad30/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L645

### DIFF
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -19,7 +19,18 @@ module Airbrake
         # Fallback mode (OmniAuth support included). Works only for Rails.
         controller = rack_env['action_controller.instance']
         return unless controller.respond_to?(:current_user)
-        new(controller.current_user) if controller.current_user
+        extract_rails_user(controller)
+      end
+
+      def self.extract_rails_user(controller)
+        return unless controller.current_user
+        new(controller.current_user)
+      ensure
+        # Don't leak DB connections.
+        # https://github.com/airbrake/airbrake/pull/562
+        if defined?(::ActiveRecord::Base)
+          ::ActiveRecord::Base.clear_active_connections!
+        end
       end
 
       def initialize(user)
--- a/spec/integration/rails/rails_spec.rb
+++ b/spec/integration/rails/rails_spec.rb
@@ -165,6 +165,21 @@ RSpec.describe "Rails integration specs" do
           /"user":{"id":"1","username":"qa-dept","email":"qa@example.com"}/
         )
       end
+
+      it "releases DB connections" do
+        user = OpenStruct.new(id: 1, email: 'qa@example.com', username: 'qa-dept')
+        allow_any_instance_of(DummyController).to receive(:current_user) { user }
+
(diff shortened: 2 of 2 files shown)


vmihailenco:
>Our Rack middleware kicks in earlier, so we have to deal with this on our own.

Did you consider changing middleware order so Rack middleware can do its job?

kyrylo:
Our middleware (and middleware of all other other similar libraries) has always been inserted after `DebugExceptions`. I cannot really tell the implications of inserting it a bit later, but here's the default Rails 4 middleware stack (with our library):

```
use Rack::Sendfile
use ActionDispatch::Static
use Rack::LiveReload
use Rack::Lock
use Rack::Runtime
use Rack::MethodOverride
use ActionDispatch::RequestId
use Rails::Rack::Logger
use ActionDispatch::ShowExceptions
use ActionDispatch::DebugExceptions
use Airbrake::Rack::Middleware
use ActionDispatch::RemoteIp
use ActionDispatch::Reloader
use ActionDispatch::Callbacks
use ActiveRecord::ConnectionAdapters::ConnectionManagement
use ActiveRecord::QueryCache
use ActionDispatch::Cookies
use ActionDispatch::Session::CookieStore
use ActionDispatch::Flash
use ActionDispatch::ParamsParser
use Rack::Head
use Rack::ConditionalGet
use Rack::ETag
use Warden::Manager
use Rack::Cors
use CustomDomainCookie
run AirbrakeApp::Application.routes
```

So, if we're taking an alternative direction, we want to insert it after `ActiveRecord::ConnectionAdapters::ConnectionManagement`. It means that errors in these errors won't be caught:

* [ActionDispatch::RemoteIp](http://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html)
* [ActionDispatch::Reloader](http://api.rubyonrails.org/classes/ActionDispatch/Reloader.html)
* [ActionDispatch::Callbacks](http://api.rubyonrails.org/classes/ActionDispatch/Callbacks.html)

I feel a bit uneasy about changing order because I cannot really predict what will happen.

vmihailenco:
Overall why we execute any SQL queries? Can we avoid it?

>I feel a bit uneasy about changing order because I cannot really predict what will happen.

Probably it will fix the issue. Probably not. Nothing dangerous.

kyrylo:
We call `current_user`, which is a common, but not standard method in all Rails apps that have user profiles. Its implementation can be anything (a library might implement it, or the app itself), so sometimes it can execute SQL queries. Let's try changing the order, since I am not convinced about the current solution, too.

PR #570, opened 2016-07-06
ignore_only=[] was removed form Airbrake 5

equivalent:
I use to use it as:

```
+  #if ENV['CONTAINER_ROLE'] == "sidekiq"
+    #config.ignore_only = []
+  #end
```

to allow raise erros from `ActiveRecord::NotFound`

by not having this in the change-log I had pretty hard time last 2 hours, would be helpful for others

### DIFF
--- a/docs/Migration_guide_from_v4_to_v5.md
+++ b/docs/Migration_guide_from_v4_to_v5.md
@@ -71,6 +71,7 @@ framework | removed | n/a
 rescue_rake_exceptions | removed | n/a
 user_attributes | removed | n/a
 test_mode | removed | n/a
+ignore_only | removed | n/a
 
 * <a name="api-key"></a>
   The `api_key` option was renamed to `project_key`.


kyrylo:
I'm not sure `ignore_only` was an Airbrake v4 option. Perhaps it was an earlier option removed in v4? If this is true, then it doesn't make sense to mention that in the v4-v5 transition doc. What was your library version?

equivalent:
it was `4.3.4` I'm pretty sure we were not using `> 4`  as this is relatively fresh project and airbrake was introduced like 6 months ago 

```
# Gemfile.lock
-    airbrake (4.3.4)
# ...
+    airbrake (5.4.1)
# ...
```

docs/Migration_guide_from_v4_to_v5.md
```diff
@@ -71,6 +71,7 @@ framework | removed | n/a
 rescue_rake_exceptions | removed | n/a
 user_attributes | removed | n/a
 test_mode | removed | n/a
+ignore_only | removed | n/a
```
kyrylo: Could you move this below `ignore_user_agent` (line 60) and instead of `removed` mention `[add_filter](#filtering)`, just like other ignore options do?

kyrylo:
Fixed by https://github.com/airbrake/airbrake/pull/576.

PR #571, opened 2016-07-07
spec_helper: fix build by requiring sinatra in its test file

kyrylo:
The bug was introduced by:
https://github.com/mperham/sidekiq/issues/3042

It resulted in loading Sinatra when we run tests against Rack (we
need Sidekiq for unit tests, so we load it).

Example build that fails:
https://circleci.com/gh/airbrake/airbrake/781


kyrylo:
Merging as a quick fix.

PR #576, opened 2016-07-15
Migration Guide: mention ignore_only being removed from Airbrake v5

kyrylo:
Alternative to #570
Thanks to: @equivalent


equivalent:
:+1:  awesome, thx

PR #580, opened 2016-07-21
rails/railtie: fix support of apps, which don't use ActiveRecord

### LINKED ISSUE #579: Middleware ConnectionManagement only for ActiveRecord
Since last version _5.4.2_, the change in *lib/airbrake/rails/railtie.rb* loads the middleware, but it should only do that for apps with ActiveRecord.

kyrylo:
Fixes #579 (Middleware ConnectionManagement only for ActiveRecord)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Airbrake Changelog
 
 ### master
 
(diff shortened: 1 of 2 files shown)


kyrylo:
PTAL. Would be nice to release the fix today.

zefer:
LGTM

PR #583, opened 2016-07-25
Always re-raise exception in ActiveJob integration

confiks:
In pull request https://github.com/airbrake/airbrake-ruby/pull/75, code was introduced in every integration to return early when `build_notice` returns nil because Airbrake is not configured.

In at least the ActiveJob integration, the exception is not re-raised when returning early, and thus any ActiveJob queue adapter will happily mark the job successfully completed. That is not the expected behavior, and it took me a while to figure out jobs were falsely successful because the Airbrake gem was included without configuring it.

This pull request contains a fix for the ActiveJob integration, but I suspect more integrations suffer from this problem.

(I also modified Rubocop's Gemfile version to exclude 0.42, because it barfed on CircleCI's MRI 1.9.3)

### DIFF
--- a/lib/airbrake/rails/active_job.rb
+++ b/lib/airbrake/rails/active_job.rb
@@ -12,19 +12,19 @@ module Airbrake
         end
 
         rescue_from(Exception) do |exception|
-          next unless (notice = Airbrake.build_notice(exception))
+          if (notice = Airbrake.build_notice(exception))
+            notice[:context][:component] = 'active_job'
+            notice[:context][:action] = self.class.name
 
-          notice[:context][:component] = 'active_job'
-          notice[:context][:action] = self.class.name
+            notice[:params] = as_json
 
-          notice[:params] = as_json
-
-          # We special case Resque because it kills our workers by forking, so
-          # we use synchronous delivery instead.
-          if is_resque_adapter
-            Airbrake.notify_sync(notice)
-          else
-            Airbrake.notify(notice)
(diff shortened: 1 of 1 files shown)


kyrylo:
This makes sense 👍  Apologies for the bug.
I've fixed the Rubocop issue by https://github.com/airbrake/airbrake/pull/584. Could you please rebase?

confiks:
Rebased. For some reason build_notice is called twice with the new logic, that's why the test is failing, but I don't immediately see why.

kyrylo:
This is expected, since `Airbrake.notify` calls it internally. If you could fix the test, I would be able to release the gem this week.

kyrylo:
Ah, sorry. This is not actually expected. I'm looking into that.

lib/airbrake/rails/active_job.rb
```diff
@@ -12,19 +12,19 @@ module ActiveJob
...
+            # We special case Resque because it kills our workers by forking, so
+            # we use synchronous delivery instead.
+            if is_resque_adapter
+              Airbrake.notify_sync(notice)
+            else
+              Airbrake.notify(notice)
+            end
           end
 
           raise exception
```
kyrylo: So the problem is that now we re-raise this and nothing catches it. That's why Airbrake tries to notify it one more time. `Airbrake.notify{_sync}` internally uses `build_notice`. Hope this makes sense. It would be nice to fix tests, so I could release the library tomorrow. If not, then no worries, I'll fix it myself.

PR #585, opened 2016-08-03
rails/railtie: use `runner` API to handle exceptions in Rails runner

kyrylo:


kyrylo:
Merging, since the user confirmed it fixed the issue for them.

PR #587, opened 2016-08-04
More specific constant name in ActiveRecord

### LINKED ISSUE #586: Updating Airbrake from 4.3.8 to 5.x interferes with application constants
* Airbrake version: 4.3.8 -> 5.0.5
* Ruby version: 2.2.2
* Framework name & version: Rails 4.1.16

#### Airbrake config

      config.project_key = 'removed_for_issue'
      config.project_id = 'removed_for_issue'
    end

#### Description

I'm seeing a strange test failure when updating our application from Airbrake 4.3.8 to 5.x (initially I went straight to 5.4.3 but when I first saw the failure I tried 5.0.5 as well and that also failed).

This is a Rails 4.1 application. We have a model named "Event" and use [enum_field](/jamesgolick/enum-field) to define a "kind" attribute. `enum_field` automatically makes a constant from the name of your enum, so it defines `Event::KINDS` as a constant containing an array of values. We have a test which expects the defined values of that constant to include certain values, e.g. `"direct"` and `"enroute"`.

However, that test fails on the update from Airbrake 4.3.8 to 5.x. `Event::KINDS` is still defined, but it has value `[:commit, :rollback]`. It looks like this is because Airbrake is now defining a `KINDS` constant on every ActiveRecord model.

Namespacing this constant would resolve the issue.

pjmorse:
By using a more specific constant name than `KINDS`, it is less likely that our constant will stomp on user model constants in ActiveRecord. Closes #586.

### DIFF
--- a/lib/airbrake/rails/active_record.rb
+++ b/lib/airbrake/rails/active_record.rb
@@ -14,7 +14,7 @@ module Airbrake
     module ActiveRecord
       ##
       # @return [Array<Symbol>] the hooks that needs fixing
-      KINDS = [:commit, :rollback].freeze
+      HOOK_KINDS = [:commit, :rollback].freeze
 
       ##
       # Patches default +run_callbacks+ with our version, which is capable of
@@ -23,7 +23,7 @@ module Airbrake
       # rubocop:disable Lint/RescueException
       def run_callbacks(kind, *args, &block)
         # Let the post process handle the exception if it's not a bugged hook.
-        return super unless KINDS.include?(kind)
+        return super unless HOOK_KINDS.include?(kind)
 
         # Handle the exception ourselves. The 'ex' exception won't be
         # propagated, therefore we must notify it here.


kyrylo:
Whoa, I was not aware that `KINDS` is defined on every model. I am not a fan of renaming, since we shouldn't expose this constant.

lib/airbrake/rails/active_record.rb
```diff
@@ -14,7 +14,7 @@ module Rails
     module ActiveRecord
       ##
       # @return [Array<Symbol>] the hooks that needs fixing
-      KINDS = [:commit, :rollback].freeze
+      HOOK_KINDS = [:commit, :rollback].freeze
```
kyrylo: Would this work for you if you made this constant private:

```rb
KINDS = [:commit, :rollback].freeze
private_constant :KINDS
```

pjmorse:
Given that the constant is only used in one place, it could be removed entirely and an array literal used where it's called. I'd be happy to update the PR with a different strategy.

pjmorse:
I'll test with the private constant.

kyrylo:
Thanks. If that doesn't work (and I suspect it doesn't, just want to test), then let's inline it :)

pjmorse:
#588 inlines the constant. Using `private_constant`, I get exceptions when enum_field tries to define a new constant with the same name as the private one.

PR #588, opened 2016-08-04
Inline array and remove constant in ActiveRecord

### LINKED ISSUE #586: Updating Airbrake from 4.3.8 to 5.x interferes with application constants

pjmorse:
To avoid stomping on user model constants in ActiveRecord, inline the value in the one place it's used, and remove it. Replaces #587, closes #586.

### DIFF
--- a/lib/airbrake/rails/active_record.rb
+++ b/lib/airbrake/rails/active_record.rb
@@ -13,17 +13,13 @@ module Airbrake
     # @see https://goo.gl/ddFNg7 Rails <4.2 implementation (bugged)
     module ActiveRecord
       ##
-      # @return [Array<Symbol>] the hooks that needs fixing
-      KINDS = [:commit, :rollback].freeze
-
-      ##
       # Patches default +run_callbacks+ with our version, which is capable of
       # notifying about exceptions.
       #
       # rubocop:disable Lint/RescueException
       def run_callbacks(kind, *args, &block)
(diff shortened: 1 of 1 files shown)


kyrylo:
👍

kyrylo:
I just released a new version with this fix (v5.4.4), thanks!

PR #591, opened 2016-08-10
Always re-raise exception in ActiveJob integration (replaces #583)

kyrylo:
Description by @confiks from https://github.com/airbrake/airbrake/pull/583:

> In pull request https://github.com/airbrake/airbrake-ruby/pull/75, code was introduced in every integration to return early when `build_notice` returns nil because Airbrake is not configured.

> In at least the ActiveJob integration, the exception is not re-raised when returning early, and thus any ActiveJob queue adapter will happily mark the job successfully completed.

> This pull request contains a fix for the ActiveJob integration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* **IMPORTANT**: fixed ActiveJob integration not re-raising exceptions, which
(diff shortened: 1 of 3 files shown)


kyrylo:
@confiks I just released the new version with this fix (v5.4.4), thanks and happy Airbraking!

PR #593, opened 2016-08-23
No SystemStackError with any active jobs

taiki45:
Hi, I got SystemStackError and found tiny improvement for this gem 💎 

Since AS's `Object#as_json` simply dump all instance variables, setting
params with `#as_json` can waste resource. Moreover, if the job has
circular reference, calling `#as_json` fails with `SystemStackError`.

We can get parameter information safely by using
`ActiveJob::Base#serialize` method and it will be enough for this
use case.

How about stopping dumping all instance variables? I didn't found any reason to do this.

### DIFF
--- a/lib/airbrake/rails/active_job.rb
+++ b/lib/airbrake/rails/active_job.rb
@@ -16,7 +16,7 @@ module Airbrake
             notice[:context][:component] = 'active_job'
             notice[:context][:action] = self.class.name
 
-            notice[:params] = as_json
+            notice[:params] = serialize
 
             # We special case Resque because it kills our workers by forking, so
             # we use synchronous delivery instead.
--- a/spec/apps/rails/dummy_app.rb
+++ b/spec/apps/rails/dummy_app.rb
@@ -58,7 +58,14 @@ if Gem::Version.new(Rails.version) >= Gem::Version.new('4.2')
   class BingoJob < ActiveJob::Base
     queue_as :bingo
 
+    class BingoWrapper
+      def initialize(bingo)
+        @bingo = bingo
+      end
+    end
+
     def perform(*_args)
+      @wrapper = BingoWrapper.new(self)
       raise AirbrakeTestError, 'active_job error'
     end
   end
--- a/spec/integration/rails/rails_spec.rb
+++ b/spec/integration/rails/rails_spec.rb
@@ -112,6 +112,16 @@ RSpec.describe "Rails integration specs" do
         ).to have_been_made.at_least_once
       end
 
+      it "does not raise SystemStackError" do
+        get '/active_job'
+        sleep 2
+
+        wait_for(
+          a_request(:post, endpoint).
+          with(body: /"type":"SystemStackError"/)
+        ).not_to have_been_made
+      end
+
       context "when Airbrake is not configured" do
         it "doesn't report errors" do
           allow(Airbrake).to receive(:build_notice).and_return(nil)


kyrylo:
Hey, @taiki45! Thanks for the PR.

Your current test is green for me even if I remove your code and leave only the test. I am not quite sure what kind of circular reference you get, so we need a test, which would prove this point.

That said, I think while it would be nice to have that kind of test, this PR could be merged even if you omit it. `serialize` indeed seems to be more appropriate 👍

taiki45:
Thank you for reviewing this!

hmm... in my local machine with Rails 4.2, this test fails. Here is full log: https://gist.github.com/taiki45/e3c53f70006cd2c4e1ecca7310d697b0

Still passing in your environment?

kyrylo:
Oh, never mind. I think I forgot to undo some code. Anyway, I confirm the test fails as indented.
Great work!

taiki45:
If the test still does not work correctly, I'm thinking omitting the test...

And just FYI, this might be minium code for SystemStackError of `as_json`:

```
~/.ghq/github.com/airbrake/airbrake no-system-stack-error M??
✗ cat a.rb
require 'active_support/all'

class BingoJob
  class BingoWrapper
    def initialize(bingo)
      @bingo = bingo
    end
  end

  def initialize
    @wrapper = BingoWrapper.new(self)
  end
end

BingoJob.new.as_json

~/.ghq/github.com/airbrake/airbrake no-system-stack-error M??
✔ bundle exec appraisal rails-4.2 ruby a.rb
>> BUNDLE_GEMFILE=/Users/taiki-ono/.ghq/github.com/airbrake/airbrake/gemfiles/rails_4.2.gemfile bundle exec ruby a.rb
/Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/instance_variables.rb:13:in `block in instance_values': stack level too deep (SystemStackError)
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/instance_variables.rb:13:in `map'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/instance_variables.rb:13:in `instance_values'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:50:in `as_json'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:159:in `block in as_json'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:159:in `each'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:159:in `map'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:159:in `as_json'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:50:in `as_json'
       	 ... 9217 levels...
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:159:in `map'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:159:in `as_json'
       	from /Users/taiki-ono/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/activesupport-4.2.7.1/lib/active_support/core_ext/object/json.rb:50:in `as_json'
       	from a.rb:15:in `<main>'
```

I noticed the SystemStackError can be caused by just assignning `self` to instance variable, but it is not similar to real world case.

taiki45:
Your response is really quick, Thanks!

kyrylo:
Haha, yeah. Going to ship this change today!

kyrylo:
Just released airbrake-5.4.5 with this PR.

PR #595, opened 2016-08-24
Airbrake v4

neb2000:


kyrylo:
Thanks for the PR. However we don't recommend using v4 for Rails 5. Please upgrade to Airbrake v5.

P.S. you might find this useful (we are going to deprecate v4 soon): https://github.com/airbrake/airbrake/issues/596

PR #603, opened 2016-09-14
Fix duplicate notifications for ActiveJob on Resque

lowang:
ActiveJob has it's own ErrorHandler which is sending exceptions, there is no need to raise Exception again in Resque error handler
Fixed README.md description how to integrate Airbrake with Resque - multiple backends are required to make failed jobs appear in Resque web panel

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,10 @@ new initializer in `config/initializers/resque.rb` with the following content:
 ```ruby
 # config/initializers/resque.rb
 require 'airbrake/resque/failure'
-Resque::Failure.backend = Resque::Failure::Airbrake
+require 'resque/failure/redis'
+require 'resque/failure/multiple'
+Resque::Failure::Multiple.classes = [Resque::Failure::Redis, Resque::Failure::Airbrake]
+Resque::Failure.backend = Resque::Failure::Multiple
 ```
 
 That's all configuration.
--- a/lib/airbrake/resque/failure.rb
+++ b/lib/airbrake/resque/failure.rb
@@ -7,12 +7,19 @@ module Resque
     # @see https://github.com/resque/resque/wiki/Failure-Backends
     class Airbrake < Base
       def save
+        return if activejob_exception?
         params = payload.merge(
           component: 'resque',
           action: payload['class'].to_s
         )
 
         ::Airbrake.notify_sync(exception, params)
+      end
+
+      # AirBrake gem already attaches error notifications to ActiveJob
+      # so it needs to be skipped or duplicate exception will be logged
+      def activejob_exception?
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -277,7 +277,10 @@ new initializer in `config/initializers/resque.rb` with the following content:
 ```ruby
 # config/initializers/resque.rb
 require 'airbrake/resque/failure'
-Resque::Failure.backend = Resque::Failure::Airbrake
+require 'resque/failure/redis'
+require 'resque/failure/multiple'
+Resque::Failure::Multiple.classes = [Resque::Failure::Redis, Resque::Failure::Airbrake]
```
kyrylo: Not sure how this change is related.

kyrylo: Oh, sorry, never mind.

lib/airbrake/resque/failure.rb
```diff
@@ -7,13 +7,20 @@ module Failure
...
           action: payload['class'].to_s
         )
 
         ::Airbrake.notify_sync(exception, params)
       end
+
+      # AirBrake gem already attaches error notifications to ActiveJob
+      # so it needs to be skipped or duplicate exception will be logged
+      def activejob_exception?
+        payload['class'].to_s == 'ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper'
```
kyrylo: I feel like using [`Module#name`](https://ruby-doc.org/core-2.3.1/Module.html#method-i-name) is a bit safer.

README.md
```diff
@@ -277,7 +277,10 @@ new initializer in `config/initializers/resque.rb` with the following content:
...
+Resque::Failure.backend = Resque::Failure::Multiple
```
kyrylo: I feel like this information shouldn't be here. We don't want to enforce Redis and we only want to inform how to configure Airbrake, so no need to mention third parties. Resque has integration instructions with multiple backends, so we're good.

lib/airbrake/resque/failure.rb
```diff
@@ -7,13 +7,20 @@ module Failure
...
+        return if activejob_exception?
         params = payload.merge(
           component: 'resque',
           action: payload['class'].to_s
         )
 
         ::Airbrake.notify_sync(exception, params)
       end
+
+      # AirBrake gem already attaches error notifications to ActiveJob
```
kyrylo: Nitpick: `Airbrake` instead of `AirBrake`.
Overall I reckon this description belongs to line 10 more, so I think the following text above line 10 will do the job:

```
Avoid duplicate errors when the error is raised inside an ActiveJob job.
```

kyrylo:
Can you set up a filter which would ignore `ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper` instead? I like this solution a bit better because it's more flexible.

PR #606, opened 2016-09-30
README: explain how to integrate Rails' filter_parameters

kyrylo:
Based on https://github.com/airbrake/airbrake-ruby/pull/108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,9 +121,9 @@ and replace `PROJECT_ID` and `PROJECT_KEY` with your values:
 rails g airbrake PROJECT_ID PROJECT_KEY
 ```
 
-[Heroku add-on][heroku-addon] users can omit specifying the key and the id and invoke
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -170,6 +170,31 @@ reporting. [The description of the API][airbrake-api] is available online.
...
+  c.blacklist_keys = Rails.application.config.filter_parameters
+end
+```
+
+There are a few important details:
+
+1. You must load `filter_parameter_logging.rb` before the Airbrake config
+2. If you use Procs to configure `filter_parameters`, you need to convert them
+   to Lambdas. Otherwise you will get `ArgumentError`
+3. If you use Procs to configure `filter_parameters`, the procs must return an
```
shifi: Nitpick: should it be `Procs` -> `Lambdas` since you mentioned in (2) they should convert to Lambdas?

PR #619, opened 2016-10-20
Ignore current_user with > 0 params

### LINKED ISSUE #618: custom `current_user` with arity > 0 raises exception
* Airbrake version: 5.5.0
* Ruby version: 2.1.10
* Framework name & version: Rails 4.2

#### Airbrake config

  config.project_key = 'xxx'
  config.project_id = 'xxx'
  config.host = 'errbit.iexposure.com'
  config.environment = Rails.env == 'production' ? Rails.root.to_s.gsub(/\/home\/rails(?:-staging)?\/(.+?)\..+\z/mi, '\1') : Rails.env
end

#### Description

If you have a `current_user` method in your controller, Airbrake will attempt to call it with 0 arguments, regardless of how many arguments are required.

This is, of course, an edge case, but there are reasons it may be valid (in our case, we have an old app that uses two completely separate sets of user models for admin and client users, so you have to tell current_user which type of user you're looking for)

The simple fix would be to add an arity check before calling current_user:

```ruby
# Note: a method with only optional arguments will have arity == -1
if [-1,0].include(controller.method(:current_user).arity)
  ... # call it, etc.
end
```

mltsy:
Fixes #618 
If there is a current_user method that requires parameters, skip calling it (to avoid crashing the app while reporting an exception)

### DIFF
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -27,6 +27,7 @@ module Airbrake
         return unless [-1, 0].include?(controller.method(:current_user).arity)
         controller.current_user
       end
+      private_class_method :try_current_user
 
       def initialize(user)
         @user = user
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -19,7 +19,7 @@ module Airbrake
         # Fallback mode (OmniAuth support included). Works only for Rails.
         controller = rack_env['action_controller.instance']
         return unless controller.respond_to?(:current_user)
-        return unless [-1,0].include(controller.method(:current_user).arity)
+        return unless [-1,0].include?(controller.method(:current_user).arity)
         new(controller.current_user) if controller.current_user
       end
 
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -18,9 +18,8 @@ module Airbrake
 
         # Fallback mode (OmniAuth support included). Works only for Rails.
         controller = rack_env['action_controller.instance']
-        return unless controller.respond_to?(:current_user)
-        return unless [-1,0].include?(controller.method(:current_user).arity)
-        new(controller.current_user) if controller.current_user
+        user = try_current_user(controller)
+        new(user) if user
       end
 
       def initialize(user)
@@ -49,6 +48,12 @@ module Airbrake
         # Try to get first and last names. If that fails, try to get just 'name'.
         name = [try_to_get(:first_name), try_to_get(:last_name)].compact.join(' ')
         name.empty? ? try_to_get(:name) : name
+      end
+      
+      def self.try_current_user(controller)
+        return unless controller.respond_to?(:current_user)
+        return unless [-1, 0].include?(controller.method(:current_user).arity)
+        return controller.current_user
       end
     end
   end
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -21,6 +21,12 @@ module Airbrake
         user = try_current_user(controller)
         new(user) if user
       end
+      
+      def self.try_current_user(controller)
+        return unless controller.respond_to?(:current_user)
+        return unless [-1, 0].include?(controller.method(:current_user).arity)
+        return controller.current_user
+      end
 
       def initialize(user)
         @user = user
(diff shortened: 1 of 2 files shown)


kyrylo:
Thanks for the PR and raising this issue!
As of now, this can't be merged because the build needs to be fixed first. One more thing that is missing is a test for this case. That said, if you feel like you don't want to do this work, just let me know and I'll come up with my own PR.

mltsy:
Ah Dang!  Github's online editor must automatically add trailing whitespace, because I could swear I removed that 3 or 4 times!  I've got a few minutes now, I'll try pulling to locally and committing from there to see if that helps.

spec/unit/rack/user_spec.rb
```diff
@@ -65,13 +66,28 @@ def env_for(url, opts = {})
...
           allow(controller).to receive(:current_user) { user }
 
           retval = described_class.extract(env)
           expect(retval).to be_a(described_class)
         end
+
+        context "but it requires parameters" do
+          before do
+            class DummyController
+              def current_user(_)
```
kyrylo: Please add tests that cover:

* `def current_user(a)`
* `def current_user(a, b)`
* `def current_user(*a)`
* `def current_user(a, *b)`
* `def current_user(a, b, *c, &d)`

spec/unit/rack/user_spec.rb
```diff
@@ -65,13 +66,28 @@ def env_for(url, opts = {})
...
-          env = env_for('/', 'action_controller.instance' => controller)
           allow(controller).to receive(:current_user) { user }
 
           retval = described_class.extract(env)
           expect(retval).to be_a(described_class)
         end
+
+        context "but it requires parameters" do
+          before do
+            class DummyController
```
kyrylo: Please use Class.new to avoid `Object.send(:remove_const, :DummyController)`.

lib/airbrake/rack/user.rb
```diff
@@ -17,9 +17,15 @@ def self.extract(rack_env)
...
         # Fallback mode (OmniAuth support included). Works only for Rails.
+        user = try_current_user(rack_env)
+        new(user) if user
+      end
+
+      def self.try_current_user(rack_env)
         controller = rack_env['action_controller.instance']
         return unless controller.respond_to?(:current_user)
-        new(controller.current_user) if controller.current_user
+        return unless [-1, 0].include?(controller.method(:current_user).arity)
```
kyrylo: Let's use the following condition:

`return if controller.method(:current_user).arity <= 0`

lib/airbrake/rack/user.rb
```diff
@@ -17,9 +17,15 @@ def self.extract(rack_env)
...
+        controller.current_user
       end
```
kyrylo: Please make this method private ([example](https://github.com/airbrake/airbrake-ruby/blob/a93cd6133f799390086735ddacbee49b8f1286a5/lib/airbrake-ruby/response.rb#L42-L51)).

mltsy:
Uh... I don't think this failure was me - something about a jruby string, and something's up with bundler... ? :)

kyrylo:
Just restarted it. It happens from time to time due to network issues.

mltsy:
Just occurred to me that perhaps the more robust way to handle this would be to just call current_user and rescue if it fails... but then I'm not sure whether that would have unintended consequences either I guess.  Just wanted to mention that in case you prefer it :wink:

lib/airbrake/rack/user.rb
kyrylo: I think the condition I suggested is wrong, but I hope you got the idea (no need to use #include and array since we can rewrite it, so it uses comparison).

 ```rb
return if controller.method(:current_user).arity != 0
```

kyrylo:
No, that would be quite inefficient. What we have now looks good, it's just that the condition can be simplified, I think.

mltsy:
Oops!  I didn't realize I had to "submit my review" before all my line comments would post - Sorry!  All those comments were form Friday - suddenly it makes sense that you didn't respond :wink:

lib/airbrake/rack/user.rb
kyrylo: Cool, thanks for the explanation. That makes sense, let's leave as is.

kyrylo:
Hah, yeah, it takes a while to get used to the new feature. Thanks for the PR, great work!

kyrylo:
I'm going to release a new version as soon as https://github.com/airbrake/airbrake/pull/624 is closed, which will probably be tomorrow.

mltsy:
Wonderful, thanks! :+1:

Issue #182, opened 2013-02-06
Airbrake sends POST body on error


mheffner:
It would be really convenient if Airbrake could send the contents of the POST body when generating an exception. In our application scenario, the JSON body has the important payload details that would help us track down why certain requests fail. I don't see a way to enable that for Airbrake?

Extra points if it could detect the posted content type (eg, application/json), and render the body in the correct formatting at the exception handler interface.

We are using airbrake version 3.1.6 with the self-hosted errbit project.

kyrylo:
Hello, everyone!

Are people still interested in this? I am happy to work on this.

If this is still useful for you guys, then I would like to clarify how everyone sees this implemented. Do you want to see a new tab in the dashboard (for example, "Raw"), that would display unparsed error reports (with syntax highlighting)?

If no-one is interested anymore, I will have to close this issue.

mheffner:
We don't use the project anymore, but IMO this is still important to have. It largely means that any request that accepts parameters in a POST body can not be replayed for testing.

kyrylo:
@mheffner thanks for the answer. Is my understanding if this issue correct?

mheffner:
@kyrylo Well, I haven't used the product in a while so I couldn't comment on the UI questions. IMO the POST bodies should be visible near by where the parameters for a GET request would be displayed.

tiagomatos:
As @mheffner stated, this is really important while debugging our API.
Would be great to visualize the body from PUT and POST requests (which hold the JSON payload).

kyrylo:
@tiagomatos let me just clarify this. Do you want to see the body of requests that caused an exception? That is, you made a POST request with some payload and it crashed your app, and the problem is that Airbrake doesn't display that (potentially) faulty body, it only says that it was a POST request. Is my understanding correct?

tiagomatos:
@kyrylo yes. When there's an exception with a POST / PUT request we can only see the params from the URL query strings but would be great if we could see the cause of the error. Which 99% of time is on the JSON payload at the request.body.

kyrylo:
Thanks! This definitely looks useful. We plan a big release of this library and I'll make sure that we support this out-of-box.

kyrylo:
@tiagomatos I just double checked everything and from my understanding we actually already support this :)

I've made this request:

```
[...]

Issue #596, opened 2016-08-24
Airbrake v4 deprecation


kyrylo:
Airbrake v4 was first released on May 27, 2014. The latest update to the library was on June 24, 2016, which was the release of v4.3.8. 

Airbrake v5 was first released on December 18, 2015. It was created to replace v4. As you may know, we are still supporting Airbrake v4, accepting and applying bug fixes, and responding to issues.

However, the time has come to move on. On November 24, 2016, we will be stopping support for Airbrake v4, this version will no longer be maintained. Realistically, there's no reason to use Airbrake v4, unless you still use Rails 2 or Ruby 1.8. Airbrake v5 supports Rails 3.2+, Ruby 1.9.2+, JRuby and even Rubinius, so please upgrade and enjoy modern experience 👯

kyrylo:
### **The support for airbrake-v4 ended today**

Thanks everybody for using Airbrake v4, I hope you upgraded to v5 by now.

grosser:
should delete this wiki now ? https://github.com/airbrake/airbrake/wiki/Customizing-your-airbrake.rb

kyrylo:
👍

PR #621, opened 2016-10-21
add http post body

halida:
#182


kyrylo:
It seems like you want to merge airbrake v4 changes to master (which is airbrake v5 now). I highly suggest you to upgrade to v5 because [v4 is going to be deprecated in 1 month](https://github.com/airbrake/airbrake/issues/596).

I am going to close this PR because in order to patch v4, you need to select the correct branch. As for the approach itself, it's hard for me to judge because I'd like to see failing payload (ideally a failing test).

PR #624, opened 2016-10-22
rack/notice_builder: attach request body to notices

### LINKED ISSUE #623: Rack: send payload of XML requests on POST
Currently, our Rack integration is capable of transmitting JSON request body. The body is being displayed in the Params panel.

The problem is that if Content-Type is XML, then this is not the case anymore. Request body is not attached anymore.

Originally reported here: https://github.com/airbrake/airbrake/issues/182#issuecomment-255278749

Reproducible with:

```
require 'rest-client'
xml = '<xml><foo>bar</foo><bar>foo</bar></xml>'
RestClient.post('http://localhost:3000/welcomes', xml, content_type: :xml)
```

Expected to see `xml` in the dashboard.
Saw nothing instead.

kyrylo:
Fixes #623 (Rack: send payload of XML requests on POST)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Airbrake Changelog
 * Fixed Rails bug with regard to the `current_user` method signature having
   parameters, while the library expected none
   ([#619](https://github.com/airbrake/airbrake/pull/619))
+* Started collecting HTTP request body for Rack compliant apps
+  ([#624](https://github.com/airbrake/airbrake/pull/624))
 
 ### [v5.6.0][v5.6.0] (October 18, 2016)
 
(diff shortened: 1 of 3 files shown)


vmihailenco:
I would limit this to 512 bytes to reduce size on the notice. Backend already uses that limit on most strings and it is likely that we will do it even more aggressively in the future.

kyrylo:
Makes sense, thanks. I updated it to 512 bytes.

PTAL

vmihailenco:
LGTM

In long-term I would like to this better organized across the notifiers, i.e. this notifier puts env.REMOTE_IP into notice.params, but request.body into notice.environment. But I am pretty sure that original idea of the notice design was that `notice.environment = env` and `notice.params = request.params + body`. And every other notifier do it in a different way...

kyrylo:
Yeah, I agree that this is not the best place and I am happy to assist with this issue.

PR #627, opened 2016-11-03
rack/notice_builder: read up to 4096 bytes from Rack request's body

### LINKED ISSUE #626: Log full HTTP post data
In issue #623 Airbrake can log raw HTTP post data now,
But it limit data size to 512 byte, Airbrake should provide option to support log full post data for better debugging.

- Add config to set HTTP post data size(current 512)
- Add config to set PAYLOAD_MAX_SIZE and MAX_NOTICE_SIZE

kyrylo:
Fixes #626 (Log full HTTP post data)

We can't really read the whole body because it will significantly
increase our average notice size. But we're trying to find the right
balance here and read more.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Read up to 4096 bytes from Rack request's body (increased from 512)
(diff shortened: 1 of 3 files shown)


vmihailenco:
LGTM

PR #632, opened 2016-11-30
gemspec: fix build by locking the public_suffix dependency

kyrylo:


kyrylo:
This is failing because of a sudden Rubocop warning. The original problem was fixed, hence merging.

PR #642, opened 2017-01-03
Prevent users from signing in during notify_airbrake

### LINKED ISSUE #641: notify_airbrake persists user login during 422 error
I'm currently handling how my application handles 422 errors (invalid authenticity token). Right now I have this...

```
class ApplicationController < ActionController::Base
  protect_from_forgery with: :exception
  rescue_from ActionController::InvalidAuthenticityToken,
              with: :rescue_422_error

  # This method is called whenever a CSRF token is invalid.
  def rescue_422_error(exception)
    # Since InvalidAuthenticityToken is not raised,
    # we need to notify Airbrake that this error has occurred.
    notify_airbrake(exception)
    redirect_to '/422'
  end
end
```

Whenever I log in with an invalid authenticity token, the user is logged in and is redirected to the /422 page. This is something I don't want to happen. No user should be logged in when they have an invalid authenticity token. What I've realized is that if I comment out `notify_airbrake(exception)`, the user will not be logged in. I believe that `notify_airbrake` is persisting the request and allowing the user to be logged in. I'm currently using Airbrake (5.3.0) and Devise (3.4.1). I have written no custom code for Devise.

thejourneydude:
### What GitHub issue(s) does this address?
Fixes #641  

### What problem does your PR solve?
If a developer calls `notify_airbrake` during a user log-in request, `notify_airbrake` will attempt to authenticate and sign in the user. This is because `notify_airbrake` calls Devise's `current_user` which attempts to sign the user in. This is not the intended behavior of `notify_airbrake`.

### What does your code solution do?
I've added a check that first calls `warden.authenticated?` to see if the user is signed in. If the user is signed in, we then call `current_user`.  
 
### Are there any special changes in the code that we should be aware of?
None.

### Anything else we should know?
For more info on how Devise and Warden are authenticating the user when calling `current_user`, view these files below:

https://github.com/plataformatec/devise/blob/v3.4.1/lib/devise/controllers/helpers.rb#L119
https://github.com/hassox/warden/blob/master/lib/warden/proxy.rb#L103
https://github.com/hassox/warden/blob/master/lib/warden/proxy.rb#L316

### DIFF
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -22,7 +22,11 @@ module Airbrake
       end
 
       def self.try_user_signed_in(rack_env)
+                puts " see me 1321321321 " * 100
+
         controller = rack_env['action_controller.instance']
+                puts " see me 65464654654 " * 100
+
         return unless controller.respond_to?(:user_signed_in?)
         puts " see me 9879879387 " * 100
         return unless [-1, 0].include?(controller.method(:user_signed_in?).arity)
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -35,7 +35,7 @@ module Airbrake
         return unless [-1, 0].include?(controller.method(:current_user).arity)
         controller.current_user
       end
-      private_class_method :try_current_user 
+      private_class_method :try_current_user
 
       def initialize(user)
         @user = user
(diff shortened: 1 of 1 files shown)


kyrylo:
Thanks! I started reviewing it, but then realised it can be fixed in a more simple way.

Replaced by: https://github.com/airbrake/airbrake/pull/643

PR #645, opened 2017-01-05
Appraisals: fix delayed_job_active_record

kyrylo:
Fixes (https://circleci.com/gh/airbrake/airbrake/985):

```
Directly inheriting from ActiveRecord::Migration is not
supported. Please specify the Rails release the migration was written
for: (StandardError)
```

The gem seems to be abandoned, so we depend on a fork here to make sure
our tests pass on rails-edge.

This is what fixed the error:
https://github.com/airbrake/delayed_job_active_record/commit/56c2b52e42a201fc0923b94ce56b28e65bd69db5


kyrylo:
Merging since this depends on https://github.com/airbrake/airbrake/pull/644 and vice versa (something has to be merged first).

PR #648, opened 2017-01-18
Update airbrake_initializer.rb.erb

hardbap:
The initializer does set defaults for `environment` and `ignore_environments'.


kyrylo:
👍

PR #650, opened 2017-01-18
notice_builder: support class definitions for builders

kyrylo:
Also:

  * add optional default builder - `RequestBodyBuilder` (appends request
    body)
  * refactor other builders, so they also use classes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,13 @@ Airbrake Changelog
 
 * **IMPORTANT:** support for Ruby 1.9.2, 1.9.3 & JRuby (1.9-mode) is dropped
   ([#646](https://github.com/airbrake/airbrake/pull/646))
-* Read up to 4096 bytes from Rack request's body (increased from 512)
-  ([#627](https://github.com/airbrake/airbrake/pull/627))
 * Fixed unwanted authentication when calling `current_user`, when Warden is
   present ([#643](https://github.com/airbrake/airbrake/pull/643))
+* Stopped collecting HTTP request bodies by default, which started happening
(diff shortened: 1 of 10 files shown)


vmihailenco:
Why we have both notice filters `airbrake.add_filter` and notice rack builders `airbrake.add_rack_builder`?

kyrylo:
Rack builders have access to the `request` object, `add_filter` can't have such access.

vmihailenco:
That answer tells me nothing and is also just not very accurate. Rack, not rack builder, has access to request. And if you want to let filters access to request you can just pass it to them

```
notice = Notice.new
notice.request = @request
@apply_filters(notice)
```

kyrylo:
I see your point. Currently, with your proposal, `notice.to_json` will contain a lot of garbage from `request`. I could try to get rid of `add_rack_builder` and use `add_filter`, however we need to design this a bit better.

What do you think of having a special key, which would hold all custom objects available to filters. When filtering is done, we remove that key from the notice, so it doesn't affect the payload.

For example:

```rb
notice = Notice.new
notice[:objects][:request] = request

  notice[:session] = notice[:objects][:request][:session]
end

def apply_filters(notice)
  # applied filters here

  notice.delete(:objects)
end
```

vmihailenco:
Looks like it already ignores unknown fields on notice - https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/notice.rb#L79-L91. But some special field sounds good too.

kyrylo:
Closed in favour of https://github.com/airbrake/airbrake/pull/654.

PR #651, opened 2017-01-20
Deprecate `Airbrake.add_rack_builder` and use `add_filter` instead

kyrylo:
`add_rack_builder` was duplicating our `add_filter` functionality, so it
was natural to merge them together. To do that, we introduced the new
API in airbrake-ruby-1.7.0 - [Notice#stash](https://github.com/airbrake/airbrake/pull/650).

More info: https://github.com/airbrake/airbrake/pull/650


airbrake.gemspec
```diff
@@ -31,7 +31,7 @@ DESC
 
   s.required_ruby_version = '>= 2.0'
 
-  s.add_dependency 'airbrake-ruby', '~> 1.6'
+  # s.add_dependency 'airbrake-ruby', '~> 1.6'
```
kyrylo: This is temporary, so I can depend on unreleased features (will be merged after airbrake-ruby is released).

PR #652, opened 2017-01-20
Refactor some integrations to stop using deprecated API

kyrylo:
Setting `component/action` like that was deprecated in
https://github.com/airbrake/airbrake-ruby/pull/151


airbrake.gemspec
```diff
@@ -31,7 +31,7 @@ DESC
 
   s.required_ruby_version = '>= 2.0'
 
-  s.add_dependency 'airbrake-ruby', '~> 1.6'
+  # s.add_dependency 'airbrake-ruby', '~> 1.6'
```
vmihailenco: revert?

PR #653, opened 2017-01-20
gemspec: bump airbrake-ruby dependency from `~> 1.6` to `~> 1.7`

kyrylo:


kyrylo:
> Should this be added to the changelog as well?

Yeah, added, thanks.

PR #658, opened 2017-01-26
generators/airbrake_initializer: mention RequestBodyFilter

kyrylo:


lib/generators/airbrake_initializer.rb.erb
```diff
@@ -60,6 +60,11 @@ Airbrake.configure do |c|
   # c.blacklist_keys = Rails.application.config.filter_parameters
 end
 
+# A filter that collects request body information. Enable it if you are sure you
+# don't send sensitive information to Airbrake in your body (such as passwords).
+# https://github.com/airbrake/airbrake#requestbodyfilter
+# Airbrake.add_filter(Airbrake::Rack::RequestBodyFilter.new)
```
shifi: Is it worth to mention here that they can define how many bytes to read?

kyrylo: Let's keep it nice and simple. There's a link where people can read more.

PR #660, opened 2017-02-01
only use Rails.version if it is defined

penguincoder:
This is an interesting situation.

I am writing a Grape API application using only Rack, however, I am using ActionMailer to send some emails in Sidekiq. Something inside one of the Action* gems is defining the module `Rails`, but it's not fully loaded, as I am not using Rails. So when Airbrake gets to the point where it wants to use the context filter, it fails because `version` is not defined in the `Rails` module.

This code just checks to see if version is a defined attribute or method of the module `Rails`.

### DIFF
--- a/lib/airbrake/rack/context_filter.rb
+++ b/lib/airbrake/rack/context_filter.rb
@@ -7,7 +7,7 @@ module Airbrake
     class ContextFilter
       def initialize
         @framework_version =
-          if defined?(::Rails)
+          if defined?(::Rails) and ::Rails.respond_to?(:version)
             "Rails/#{::Rails.version}"
           elsif defined?(::Sinatra)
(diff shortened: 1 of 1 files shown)


lib/airbrake/rack/context_filter.rb
```diff
@@ -7,7 +7,7 @@ module Rack
     class ContextFilter
       def initialize
         @framework_version =
-          if defined?(::Rails)
+          if defined?(::Rails) and ::Rails.respond_to?(:version)
```
kyrylo: Rubocop will probably reject `and`. We use `&&` in this project.

kyrylo:
For some reason Circle didn't run tests for this PR. Could you fix https://github.com/airbrake/airbrake/pull/660#discussion_r99092759, it will restart the tests.

PR #664, opened 2017-02-08
Add additional check for not existing notice

incubus:
Fix exception when notifier is not configurated.

### DIFF
--- a/lib/airbrake/resque/failure.rb
+++ b/lib/airbrake/resque/failure.rb
@@ -7,7 +7,7 @@ module Resque
     # @see https://github.com/resque/resque/wiki/Failure-Backends
     class Airbrake < Base
       def save
-        notice = ::Airbrake.build_notice(exception, payload)
+        notice = ::Airbrake.build_notice(exception, payload) or return
         notice[:context][:component] = 'resque'
         notice[:context][:action] = payload['class'].to_s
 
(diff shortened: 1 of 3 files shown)


kyrylo:
Could you describe the scenario where you experienced this? The motivation of this change is currently a bit mysterious.

incubus:
Hi!

Disabled airbrake configuration in development mode, for example. You already have this checks in the code: 

https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rake/task_ext.rb#L29
https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rails/action_controller.rb#L33
https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rails/active_job.rb#L15

and so on.

Thanks.

kyrylo:
Could you add test(s) similar to https://github.com/airbrake/airbrake/commit/259d69a9b5dc796ea89ca12171986e1ba3fd3efc#diff-7a0b88384477299d930b938378611dfcR105

We also need to re-trigger Circle CI because for some reason building pull requests was disabled. I think there will be a few Rubocop offences.

lib/airbrake/resque/failure.rb
```diff
@@ -7,7 +7,7 @@ module Failure
     # @see https://github.com/resque/resque/wiki/Failure-Backends
     class Airbrake < Base
       def save
-        notice = ::Airbrake.build_notice(exception, payload)
+        notice = ::Airbrake.build_notice(exception, payload) or return
```
kyrylo: Nitpick: we use `return unless (notice = ::Airbrake.build_notice(exception, payload))`

kyrylo:
Closed in favour of https://github.com/airbrake/airbrake/pull/665.

PR #665, opened 2017-02-08
Add additional check for not existing notice

incubus:
With only one additional test for sidekiq, sorry.


kyrylo:
Thanks, this looks great!

PR #672, opened 2017-02-22
generators/airbrake_initializer: ignore HTTP_AUTHORIZATION header

### LINKED ISSUE #671: Do not log HTTP_AUTHORIZATION header by default
Same as 'password' afaik ... so please don't send it by default ...

kyrylo:
Fixes #671 (Do not log HTTP_AUTHORIZATION header by default)


grosser:
👍

PR #674, opened 2017-02-23
Add Logger integration

kyrylo:


spec/unit/logger/logger_ext_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  context "when Airbrake is installed" do
+    let(:out) { StringIO.new }
+    let(:logger) { Logger.new(out) }
+
+    before do
+      logger.airbrake = airbrake
+    end
+
+    it "both logs and notifies with the correct severity" do
```
vmihailenco: "with the correct severity" - looks like severity is tested below

README.md
```diff
@@ -412,6 +413,35 @@ task :foo do
...
+
+To start using the integration create a new Logger and assign an Airbrake
+notifier to it:
+
+```ruby
+logger = Logger.new(STDOUT)
+
+# New method.
+# We assign the default notifier here.
+logger.airbrake = Airbrake[:default]
```
vmihailenco: I would expect that `logger.airbrake == Airbrake[:default]` be default if you import `airbrake/logger/logger_ext`.

README.md
```diff
@@ -412,6 +413,35 @@ task :foo do
...
 
+### Logger
+
+If you want to convert your log messages to Airbrake errors, you can use our
+integration with Ruby's `Logger` class from stdlib. All you need to do is to
+make sure that the integration is required (if the Airbrake is loaded after
+`Logger`, then it is already integrated):
+
+```ruby
+require 'airbrake/logger/logger_ext'
```
vmihailenco: Why not simply `airbrake/logger` or `airbrake/logger_ext`?

README.md
```diff
@@ -412,6 +413,35 @@ task :foo do
...
+# New method.
+# We assign the default notifier here.
+logger.airbrake = Airbrake[:default]
+```
+
+Then, simply use the logger as you normally do. Messages will be both logged
+and sent to Airbrake.
+
+```ruby
+logger.fatal('oops')
```
vmihailenco: I wonder what about default Rails logger? Should user change all his code to replace `rails_logger` with `airbrake_logger`?

vmihailenco:
I think we should support setting severity threshold to not report `debug` and `info` calls to airbrake, e.g. `logger.AirbrakeSeverity` or `logger.AirbrakeSeverityThreshold`. This can be done in filter, but it is very common and performance will be better if you do it in logger without creating a `notice`.

vmihailenco:
Also this probably should live in airbrake-ruby since Logger is in standard library.

lib/airbrake/logger/logger_ext.rb
```diff
@@ -0,0 +1,58 @@
...
+  private
+
+  def notify_airbrake(message, severity)
+    notice = Airbrake.build_notice(message || progname)
+
+    # Get rid of unwanted internal Logger frames.
+    # Example: /ruby-2.4.0/lib/ruby/2.4.0/logger.rb
+    notice[:errors].first[:backtrace].shift
+
+    notice[:context][:component] = 'logger'
```
vmihailenco: Makes sense. Probably just `log` instead of `logger` and use the same name everywhere? In Python it is called `logging`, in Go `log`.

README.md
kyrylo: To conform the general structure that we have in this project and to be able to add more files related to logger in the same directory.

README.md
kyrylo: I had tried this when I implemented it first. This will lead to a lot of spam and unpredicted results. Many libraries use Logger including airbrake-ruby. If we do what you suggest, then Airbrake is getting bombarded by requests, including the ones that we emit during `notify` calls and including from all those libraries that use Logger. This is very unpredictable.

README.md
kyrylo: Explained above.

lib/airbrake/logger/logger_ext.rb
kyrylo: Changed.

README.md
kyrylo: I guess that's why you suggest severityThreshold.

kyrylo:
> I think we should support setting severity threshold to not report debug and info calls to airbrake, 

Implemented.

> Also this probably should live in airbrake-ruby since Logger is in standard library.

I feel like this is the correct place because we keep all integrations here. We already have the Rake integration here, which is also in stdlib.

PTAL

spec/unit/logger/logger_ext_spec.rb
kyrylo: Not sure what you mean.

kyrylo: Ah, understood. Fixed.

lib/airbrake/logger/logger_ext.rb
```diff
@@ -0,0 +1,78 @@
...
+    # * /ruby-2.4.0/lib/ruby/2.4.0/logger.rb
+    # * /gems/activesupport-4.2.7.1/lib/active_support/logger.rb
+    backtrace = notice[:errors].first[:backtrace]
+    notice[:errors].first[:backtrace] =
+      backtrace.drop_while { |frame| frame[:file] =~ %r{/logger.rb\z} }
+
+    notice[:context][:component] = 'log'
+    notice[:context][:severity] = airbrake_severity(severity)
+
+    Airbrake.notify(notice)
```
vmihailenco: Looks like ignore `Logger.airbrake`

kyrylo: Could you elaborate? I'm not sure what you mean.

vmihailenco: Looks like this code uses global Airbrake notifier, not local `logger.airbrake`. For clarity would be nice to rename `logger.airbrake` to `logger.notifier`.

kyrylo: Uh, of course, it's a mistake indeed. What about `logger.airbrake_notifier`? We monkey-patch a class and it's better to be verbose.

wlipa:
Having Logger monkey patched and behavior changed just on the basis of the Logger class existing seems like a really bad idea, since almost everything uses Logger.  This should be an opt-in behavior on Airbrake's part.

kyrylo:
Thanks for sharing your concerns. What exactly is bad about it in your opinion?

wlipa:
Monkey patching stdlib classes is a bit of a code smell due to fragility and interaction concerns.  This kind of modification of Logger is not something I would expect Airbrake to do just by including the gem.  Not to mention, having log messages trigger Airbrake is a big change.  It's something people should opt into to avoid unwanted surprises.

I would much prefer it to be something that has to be turned on by setting an Airbrake config option or wrapping a Logger with AirbrakeLogger the way Rails TaggedLogger works.

kyrylo:
Thanks!

Sometimes monkey-patching works quite nicely (see our Rake integration).

> I would much prefer it to be something that has to be turned on by setting an Airbrake config option or wrapping a Logger with AirbrakeLogger the way Rails TaggedLogger works.

😄  https://github.com/airbrake/airbrake/pull/685

wlipa:
Much better!

PR #683, opened 2017-02-28
logger: improve support

kyrylo:
* I'm still a bit wary to enable this integration by default because I
  don't know what to expect from this. One of our
  users also has some concerns:
  https://github.com/airbrake/airbrake/pull/674#issuecomment-282953767

  Thus, I want to keep this as an optional integration for now. If this
  proves to be useful and harmless, I'm ready to enable it by default

* The code here is a bit funky now due to the fact that there was a bug
  in the initialize method. Loggers initialized before Airbrake is
  included (Rails initializes at least 4 loggers before Airbrake is
  loaded) don't have a notifier set.

* Severity cannot be set to DEBUG or INFO to avoid massive spam. When that happens our own airbrake-ruby spams the dashboard with requests continuously, which leads to DDoS'ing our service. We can technically stop logging every response in airbrake-ruby, but this is still very dangerous because any other library that we do not control can do the same without even realising it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Fixed Logger integration error and refactored the way it works. It's no longer
+  required automatically ([#681](https://github.com/airbrake/airbrake/pull/681))
+
 ### [v5.8.0.rc.2][v5.8.0.rc.2] (February 27, 2017)
 
 * Fixed Rails controller helper methods throwing `NameError`
--- a/lib/airbrake/logger/logger_ext.rb
+++ b/lib/airbrake/logger/logger_ext.rb
@@ -14,30 +14,30 @@
 class Logger
   # Store the orginal methods to use them later.
   alias add_without_airbrake add
-  alias initialize_without_airbrake initialize
-
-  ##
-  # @see https://goo.gl/MvlYq3 Logger#initialize
-  def initialize(*args)
-    @airbrake_notifier = Airbrake[:default]
-    @airbrake_severity_level = WARN
-    initialize_without_airbrake(*args)
-  end
 
   ##
   # @return [Airbrake::Notifier] notifier to be used to send notices
   attr_accessor :airbrake_notifier
 
   ##
+  # @return [Integer]
+  attr_reader :airbrake_severity_level
+
+  ##
+  # Sets airbrake severity level. Does not permit values below `Logger::WARN`.
+  #
   # @example
   #   logger.airbrake_severity_level = Logger::FATAL
-  # @return [Integer]
-  attr_accessor :airbrake_severity_level
+  # @return [void]
+  def airbrake_severity_level=(level)
+    level = WARN if level < WARN
+    @airbrake_severity_level = level
+  end
 
   ##
   # @see https://goo.gl/8zPyoM Logger#add
   def add(severity, message = nil, progname = nil, &block)
-    if severity >= airbrake_severity_level && airbrake_notifier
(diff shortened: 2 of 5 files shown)


lib/airbrake/logger/logger_ext.rb
```diff
@@ -75,4 +75,20 @@ def airbrake_severity(severity)
...
+
+  ##
+  # @macro helper_method
+  def current_airbrake_notifier
+    airbrake_notifier || Airbrake[:default]
+  end
+
+  ##
+  # @marcro helper_method
+  def current_airbrake_severity
```
vmihailenco: `current_` prefix is misleading. It is usually used when var changes during method execution, e.g. in iteration. It is also not clear why `airbrake_severity_level` becomes `current_airbrake_severity`.

lib/airbrake/logger/logger_ext.rb
```diff
@@ -14,30 +14,30 @@
...
+    level = WARN if level < WARN
+    @airbrake_severity_level = level
+  end
 
   ##
   # @see https://goo.gl/8zPyoM Logger#add
   def add(severity, message = nil, progname = nil, &block)
-    if severity >= airbrake_severity_level && airbrake_notifier
+    if severity >= current_airbrake_severity && current_airbrake_notifier
       notify_airbrake(severity, message || progname)
```
vmihailenco: It is confusing that you check for `current_airbrake_notifier`, but don't use it.

lib/airbrake/logger/logger_ext.rb
kyrylo: I could not come up with a better name. These shenanigans are needed to eliminate ruby warnings about unassigned instance variables that used to be in the initialise method.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Fixed Logger integration error and refactored the way it works. It's no longer
```
vmihailenco: `Fixed Logger integration error and refactored the way it works` provides 0 information.

lib/airbrake/logger/logger_ext.rb
```diff
@@ -14,30 +14,30 @@
...
+  ##
+  # Sets airbrake severity level. Does not permit values below `Logger::WARN`.
+  #
   # @example
   #   logger.airbrake_severity_level = Logger::FATAL
-  # @return [Integer]
-  attr_accessor :airbrake_severity_level
+  # @return [void]
+  def airbrake_severity_level=(level)
+    level = WARN if level < WARN
```
vmihailenco: why not to throw an exception?

vmihailenco:
I did not realize `class Logger` does monkey patching by default. Any chance you can create a decorator:

```
logger = new Logger()
logger = airbrakenize(notifier, logger)
```

?

kyrylo:
> Any chance you can create a decorator

Yes, this sounds less hacky to me.

lib/airbrake/logger/logger_ext.rb
vmihailenco: The classic way is to use `get_`. Or rename `airbrake_severity_level` -> `_airbrake_severity_level`, `current_airbrake_severity` -> `airbrake_severity_level`

vmihailenco:
An alternative would be to add functionality to enable or disable this per file/module/package, e.g.:

```
logger = new Logger()
logger.enable_for_path('models/*')
logger.enable_for_path('models/some.rb')
```

and logger will have to check backtrace for every log call (which can be slow)...

kyrylo:
Closed in favour of https://github.com/airbrake/airbrake/pull/685.

PR #707, opened 2017-04-01
Place all the dependencies required for Rack support in one file

### LINKED ISSUE #706: If you load Airbrake before Rack is defined it's troublesome to use its Rack support afterward
Airbrake version: 6.0.0

#### Description

We have a situation where we need to load Rack support for Airbrake at a later time. However, to do that we currently have to include all [these](https://github.com/airbrake/airbrake/blob/v6.0.0/lib/airbrake.rb#L12-L18) files. It would be easier if we just had one file, say `airbrake/rack`, that we needed to import.

dwayne:
Closes #706

### DIFF
--- a/lib/airbrake.rb
+++ b/lib/airbrake.rb
@@ -9,13 +9,7 @@ require 'airbrake/version'
 
 # Automatically load needed files for the environment the library is running in.
 if defined?(Rack)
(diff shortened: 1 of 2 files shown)


kyrylo:
This makes sense. Thanks!

PR #710, opened 2017-04-06
Release v4.3.9

kyrylo:
This is an unscheduled support release. All v4 users *must* upgrade to
this version to be able to send errors OR configure the `secure` option:

```rb
  config.secure = true
end
```

Due to some changes in our infrastructure, we don't accept HTTP payload anymore and our older
notifiers are configured to send errors via HTTP by default.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,9 @@
+Version 4.3.9 - 2017-04-06 13:39:44 +0300
+===============================================================================
+
+Kyrylo Silin (1):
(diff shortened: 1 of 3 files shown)


kyrylo:
We decided to postpone this decision for some time. I still recommend all airbrake v4 users to *upgrade* as soon as possible because v4 is unsupported (as per https://github.com/airbrake/airbrake/issues/596).

If you continue using v4, make sure to enable the `secure` option to protect your data.

PR #712, opened 2017-04-10
README.md : fix Airbrake.add_filter example

AntoineBecquet:
Just came across this error.

Is that ok as a pull request ?


kyrylo:
Thanks for fixing 👍

PR #716, opened 2017-04-24
Fix Rack integration overriding 'notice[:params]'

### LINKED ISSUE #715: [BUG] HttpParamsFilter overrides custom params.
When putting extra params in an exception using `to_airbrake` or when using the `add_filter` like so:
```ruby
  notice[:params][:xxxx] = "1234"
end
```
They get overriden if the Airbrake notice contains the rack request in filtering (controller errors, etc) here: https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/notifier.rb#L124
Because the custom filter can be applied before `HttpParamsFilter`

Therefore this line here: https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rack/http_params_filter.rb#L13
```ruby
notice[:params] = request.params
```
needs to be:
```ruby
notice[:params].merge! request.params
```

kyrylo:
Fixes #715 (HttpParamsFilter overrides custom params)

### DIFF
--- a/lib/airbrake/rack/http_params_filter.rb
+++ b/lib/airbrake/rack/http_params_filter.rb
@@ -10,7 +10,7 @@ module Airbrake
       def call(notice)
         return unless (request = notice.stash[:rack_request])
 
-        notice[:params] = request.params
+        notice[:params].merge!(request.params)
 
         rails_params = request.env['action_dispatch.request.parameters']
         notice[:params].merge!(rails_params) if rails_params
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Airbrake Changelog
 * Started appending HTTP request info (method, headers & referer) from Rack to
   the `context` hash instead of the `environment` hash
   ([#703](https://github.com/airbrake/airbrake/pull/703))
+* Fixed Rack integration overriding `notice[:params]`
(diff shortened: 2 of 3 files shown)


lib/airbrake/rack/http_params_filter.rb
```diff
@@ -10,7 +10,7 @@ class HttpParamsFilter
       def call(notice)
         return unless (request = notice.stash[:rack_request])
 
-        notice[:params] = request.params
+        notice[:params].merge!(request.params)
```
PeterWuMC: I would imagine you want it to be reverse merge.
so your filter can mask the request params from sensitive information

vmihailenco:
Why not to save them under 2 separate keys?

lib/airbrake/rack/http_params_filter.rb
dragosmiron: i think there is a separate filter that does the masking?

kyrylo: Yeah, we have a filter that does this dirty job, so we should be good here.

kyrylo:
> @vmihailenco 
> Why not to save them under 2 separate keys?

`notice[:params][:httpParams]`?

kyrylo:
Merging as is for now.

PR #717, opened 2017-04-27
Add respository to install_heroku_deploy_hook task

sgray:


lib/airbrake/rake/tasks.rb
```diff
@@ -106,8 +106,19 @@
            " environment will be used."
     end
 
+    unless (repo = ENV['REPOSITORY_URL'])
+      repo = `git remote get-url origin`.chomp
+      if !repo.empty?
```
kyrylo: `if repo.empty?` with swapped bodies will help to get rid of the `!`.

lib/airbrake/rake/tasks.rb
```diff
@@ -106,8 +106,19 @@
            " environment will be used."
     end
 
+    unless (repo = ENV['REPOSITORY_URL'])
+      repo = `git remote get-url origin`.chomp
```
kyrylo: This will print `fatal: No such remote 'origin'` to stderr when `origin` is undefined. Should be something like this instead:

```ruby
 repo = `git remote get-url origin 2>/dev/null`.chomp

File: lib/airbrake/rake/tasks.rb (hunk 1 above)
Comment (sgray) [inline lib/airbrake/rake/tasks.rb]: Right, but I think it reads better this way with the `else` block containing the message when neither of our first two checks worked.

File: lib/airbrake/rake/tasks.rb (hunk 2 above)
Comment (sgray) [inline lib/airbrake/rake/tasks.rb]: 👍

File: lib/airbrake/rake/tasks.rb (hunk 1 above)
Comment (kyrylo) [inline lib/airbrake/rake/tasks.rb]: `if !` never reads better in Ruby :) Then we could use `unless`.

Comment (sgray) [inline lib/airbrake/rake/tasks.rb]: But I don't want to use `unless` with `else`. 

>`if !` never reads better in Ruby :)

I disagree in this case, but I don't really mind. I'll swap the messages.

File: lib/airbrake/rake/tasks.rb [hunk 3]
```diff
@@ -106,8 +106,19 @@
            " environment will be used."
     end
 
+    unless (repo = ENV['REPOSITORY_URL'])
+      repo = `git remote get-url origin 2>/dev/null`.chomp
+      if repo.empty?
+        puts "Airbrake couldn't identify your app's repository."
```
kyrylo: I think this message can be more concrete: ``Airbrake couldn't identify `origin` of your app's repository`` (or something similar). This way it's obvious why it couldn't detect the repo.

sgray: I was thinking it should be more generic because it means two checks have failed, they didn't set the `REPOSITORY_URL` env var _and_ we couldn't get the url for the `origin` remote.

kyrylo: How will users discover that `REPOSITORY_URL` is a thing?

sgray: I'll add a PR to update the documentation. The heroku deploy docs are in need of an update anyway, so I'll include this change along with the rest.

PR #721, opened 2017-04-28
rake/tasks: don't load initializer when Airbrake is available

### LINKED ISSUE #714: do not load airbrake initializer when environment is already loaded and require it
- do not load when it was already loaded via then environment -> check if environment task was already invoked
 - use `require File.expand_path(file)` instead of load load prevent duplicate loading (in case environment is loaded later)

https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rake/tasks.rb#L71

kyrylo:
Fixes #714 (do not load airbrake initializer when environment is already
loaded and require it)


grosser:
why not use require ?

kyrylo:
Didn't work.

PR #730, opened 2017-05-05
Release v6.1.0.rc.1

kyrylo:


kyrylo:
I noticed a lot of failing specs, but Circle reported that the build is green. Needs investigation.

kyrylo:
Blood, sweat & tears... but it's all fixed now.

PR #746, opened 2017-05-22
rake: check for NilNotifier

### LINKED ISSUE #745: Capistrano Deployment Notification stopped working in v6.1.0 Update
We recently updated to v6.1.0 and deployment notifications for Capistrano v3 have stopped working.  Rolling back just the `airbrake` gem to v6.0.0 resolves the issue.

kyrylo:
Fixes #745 (Capistrano Deployment Notification stopped working in v6.1.0 Update)

`Airbrake::NilNotifier` was introduced in airbrake-ruby v2.1.0. This
check wasn't updated accordingly.

### DIFF
--- a/lib/airbrake/rake/tasks.rb
+++ b/lib/airbrake/rake/tasks.rb
@@ -67,8 +67,8 @@ namespace :airbrake do
 
       # Avoid loading the environment to speed up the deploy task and try guess
       # the initializer file location.
-      if initializer.exist?
-        load(initializer) unless Airbrake[:default]
+      if initializer.exist? && Airbrake[:default].is_a?(Airbrake::NilNotifier)
+        load(initializer)
       else
         Rake::Task[:environment].invoke
       end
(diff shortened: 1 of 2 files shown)


lib/airbrake/rake/tasks.rb
```diff
@@ -67,8 +67,8 @@
 
       # Avoid loading the environment to speed up the deploy task and try guess
       # the initializer file location.
-      if initializer.exist?
-        load(initializer) unless Airbrake[:default]
+      if initializer.exist? && Airbrake[:default].is_a?(Airbrake::NilNotifier)
```
vmihailenco: 1. Check for NilNotifier is hacky
2. I have no idea why you load initialized only when default notifier is NilNotifier

kyrylo: > Check for NilNotifier is hacky

It is. This whole expression is a hack to make the task run faster.

> I have no idea why you load initialized only when default notifier is NilNotifier

That means Airbrake wasn't configured yet. To send deploy info, we need a configured notifier.

This hack is needed to speedup the task. It used to works quite reliably, if we don't count the fact that it got broken in this release :)

PR #747, opened 2017-05-23
rake/tasks: improve feedback of 'airbrake:deploy'

kyrylo:
Depends on: https://github.com/airbrake/airbrake-ruby/pull/226

We can leverage promises to display user friendly messages and return
proper exit codes.

### DIFF
--- a/lib/airbrake/rake/tasks.rb
+++ b/lib/airbrake/rake/tasks.rb
@@ -74,13 +74,18 @@ namespace :airbrake do
       end
     end
 
-    Airbrake.create_deploy(
+    deploy_params = {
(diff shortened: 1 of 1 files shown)


vmihailenco:
Does it work as is? I expect to see some mechanism that waits for promise to be resolved or rejected. I guess this is done with `at_exit`, but something more obvious would be welcome.

kyrylo:
It works as is. This change is purely cosmetic.
The proper mechanism would be calling `promise.value`, but in our case it's non-blocking.
Yes, `at_exit` makes sure the task doesn't exit prematurely.

PR #755, opened 2017-06-09
Don't call methods of @user if parameters are required

valdis:
If attribute methods are overwritten to have parameters
e.g.
```
class User < ApplicationRecord
  def name(formating)
  end
end
```
Airbrake will fail with error `ArgumentError (wrong number of arguments (given 0, expected 1))`
and no error message will be sent to Airbrake

### DIFF
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -50,7 +50,11 @@ module Airbrake
       private
 
       def try_to_get(key)
-        String(@user.__send__(key)) if @user.respond_to?(key)
+        return unless @user.respond_to?(key)
+        # try methods with no arguments or with variable number of arguments,
+        # where none of them are required
+        return unless @user.method(key).arity.between?(-1, 0)
+        String(@user.__send__(key))
       end
 
       def full_name
--- a/spec/unit/rack/user_spec.rb
+++ b/spec/unit/rack/user_spec.rb
@@ -189,5 +189,34 @@ RSpec.describe Airbrake::Rack::User do
         expect(user_data).to be_empty
       end
     end
+
+    context "when Rack user's field expects a parameter" do
+      let(:user_data) do
+        described_class.new(
+          Class.new do
+            def name(required_param)
+              "my name is #{required_param}"
+            end
+
+            def id(required_param, *optional_params)
+              "id is #{required_param} #{optional_params.inspect}"
+            end
+
+            def username(*optional_params)
+              "username is #{optional_params.inspect}"
+            end
+          end.new
+        ).as_json
+      end
+
+      it "does not call the method if it has required parameters" do
+        expect(user_data[:user]).not_to include(:id)
+        expect(user_data[:user]).not_to include(:name)
+      end
+
+      it "calls the method if it has a variable number of optional parameters" do
+        expect(user_data[:user]).to include(:username)
+      end
+    end
   end
 end


kyrylo:
Looks good, but it needs:

1. Tests
2. Green build
3. A sip of hot tea/coffee after the job is done 😁

valdis:
Should be good now

lib/airbrake/rack/user.rb
```diff
@@ -50,7 +50,9 @@ def as_json
       private
 
       def try_to_get(key)
-        String(@user.__send__(key)) if @user.respond_to?(key)
+        String(@user.__send__(key)) if @user.respond_to?(key) &&
+                                       (@user.method(key).arity.zero? ||
```
xymbol: Would extract a method here. Then, do:

```ruby
def can_call?(sym)
  @user.respond_to?(sym) && @user.method(sym).arity <= 0
end
```

spec/unit/rack/user_spec.rb
```diff
@@ -189,5 +189,19 @@ def current_user
         expect(user_data).to be_empty
       end
     end
+
+    context "when Rack user field getters require parameters" do
+      class UserRequiringParameters
```
xymbol: Would probably rephrase to something like "... user has methods with arguments". Then, would use a `let` block and an anonymous class as in other examples.

spec/unit/rack/user_spec.rb
```diff
@@ -189,5 +189,19 @@ def current_user
...
+        def name(hello)
+          "my name is #{hello}"
+        end
+      end
+
+      let(:user_data) { described_class.new(UserRequiringParameters.new).as_json }
+
+      it "is empty" do
```
xymbol: Would rephrase to make the intention more explicit. For example, "does not call the methods with arguments".

lib/airbrake/rack/user.rb
```diff
@@ -50,7 +50,9 @@ def as_json
...
+                                        @user.method(key).arity == -1)
```
xymbol: Would extract a method and do:

```ruby
def can_call?(sym)
  @user.respond_to?(sym) && @user.method(sym).arity <= 0
end
```

valdis: comparison with <= 0 will not work, because methods with variable number of arguments, that have also required arguments will return < -1

valdis:
third time's the charm :)

lib/airbrake/rack/user.rb
```diff
@@ -50,7 +50,12 @@ def as_json
       private
 
       def try_to_get(key)
-        String(@user.__send__(key)) if @user.respond_to?(key)
+        String(@user.__send__(key)) if can_call_method?(key)
```
kyrylo: My preference would be the following implementation, however I appreciate the feedback from the previous reviewer. ⛳️
 
```rb
def try_to_get(key)
  return unless @user.respond_to?(key)
  return if (-1..0).include?(@user.method(key).arity)
  String(@user.__send__(key))
end

File: spec/unit/rack/user_spec.rb [hunk 6]
```diff
@@ -189,5 +189,21 @@ def current_user
         expect(user_data).to be_empty
       end
     end
+
+    context "when Rack user class has methods with arguments" do
```
kyrylo: This is a bit cryptic. How about:

> when Rack user defines a field with a parameters

kyrylo: Corrected my example

spec/unit/rack/user_spec.rb
```diff
@@ -189,5 +189,21 @@ def current_user
...
+        described_class.new(
+          Class.new do
+            def name(hello)
+              "my name is #{hello}"
+            end
+          end.new
+        ).as_json
+      end
+
+      it "does not call the method if it has arguments" do
```
kyrylo: Continuing from the previous example:

> when Rack user's field expects a parameter
> it skips field invocation

kyrylo:
> note that arity comparison with <= 0 will not work, because methods with variable number of arguments, that have also required arguments will return < -1

It would be nice to have a test for that.

xymbol:
>note that arity comparison with <= 0 will not work, because methods with variable number of arguments, that have also required arguments will return < -1

You're right. Still, I would avoid the double introspection:

```ruby
# using between?
@user.method(name).arity.between? 0, -1

# using include?
[0, 1].include? @user.method(name).arity

# explicit
arity = @user.method(name).arity
arity == 0 || arity == -1
```

Options, options.

lib/airbrake/rack/user.rb
xymbol: Didn't see this, suggested something similar below.

valdis: guys, this is not cool, i don't have neither the time nor patience to change the code up time and time again, until all of you like it. Discuss first internally and then give suggestions.

kyrylo: I'm sorry for the confusion @valdis, however I'd like to make it clear that @xymbol doesn't maintain this repo and he's not our employee.

You can see that my badge says "Owner", and @xymbol's doesn't.

don t call methods of user if parameters are required by valdis pull request 755 airbrake airbrake 2017-06-13 16-03-07

kyrylo: @valdis do you want me to take over this PR? Happy to help.

xymbol: @valdis Sorry to read this. My comment was well-intentioned and, it's not a matter of aesthetic taste. Anything that goes into a library of this kind gets called over and over and eventually does affect performance. Peer review is good.

kyrylo:
Forgot about the existence of `#between?`. Cool trick!

valdis:
i hope this is good enough now :)

kyrylo:
Thanks for the gotcha. Given that this is deprecated and overriding those methods isn't super common (first time I see this), I think we can ignore this issue with required params for now.

Issue #754, opened 2017-06-05
Append context/userAddr by default


kyrylo:
All Rack integrations should append this value by default.

PR #756, opened 2017-06-09
Append userAddr to context

xymbol:
Checked with Airbrake today and learned that user addresses were not being included in notices. This is particularly confusing on Heroku, as only AWS addresses are reported.

This adds `userAddr` to notices and relies on Rack's well-tested `#ip` method. https://github.com/rack/rack/blob/c1437097dcdf92d53a692ca8135a3391791fbca3/lib/rack/request.rb#L256-L265

### DIFF
--- a/spec/unit/rack/context_filter_spec.rb
+++ b/spec/unit/rack/context_filter_spec.rb
@@ -44,6 +44,28 @@ RSpec.describe Airbrake::Rack::ContextFilter do
     end
   end
 
+  context "when visitor address is present" do
+    let(:opts) do
+      { 'REMOTE_ADDR' => '1.2.3.4' }
+    end
+
+    it "adds userAddr to the context" do
+      subject.call(notice)
+      expect(notice[:context][:userAddr]).to eq('1.2.3.4')
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


spec/unit/rack/context_filter_spec.rb
```diff
@@ -38,7 +38,7 @@ def env_for(url, opts = {})
       { 'HTTP_USER_AGENT' => 'Bingo Agent' }
     end
 
-    it "adds User-Agent to the context" do
+    it "adds UserAgent to the context" do
```
kyrylo: Could you please revert this? It's not a typo. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

kyrylo:
This is super cool! Will fix https://github.com/airbrake/airbrake/issues/754.

spec/unit/rack/context_filter_spec.rb
xymbol: My bad, assumed the test referred to the `userAgent` data field in the API. https://airbrake.io/docs/api/#create-notice-v3

PR #763, opened 2017-06-30
rake: fix the deploy and test tasks for plain ruby projects

### LINKED ISSUE #762: `rake airbrake:deploy` doesn't seem to work in a non-Rails project
* Airbrake version: 6.1.2 / 2.2.7
* Ruby version: 2.4.1p111
* Framework name & version: None

#### Airbrake config

N/A, but see below. I haven't actually set it up in this test, but it fails before it gets to that point.

#### Description

I'm trying to follow the docs for notifying Airbrake of a deployment, specifically the [using-the-rake-task "Anywhere" part](https://github.com/airbrake/airbrake#user-content-rake-task).

I've got the following very minimal test case:
```
$ ls -l
total 12
-rw-r--r-- 1 nick nick  46 Jun 28 14:48 Gemfile
-rw-r--r-- 1 nick nick 198 Jun 28 14:48 Gemfile.lock
-rw-r--r-- 1 nick nick  54 Jun 28 14:46 Rakefile
```

The `Gemfile` only contains `airbrake`:
```
$ cat Gemfile
source "https://rubygems.org"

gem "airbrake"
```

`Rakefile` includes the Rake integration, as indicated in the docs:
```
$ cat Rakefile
require 'bundler/setup'
require 'airbrake/rake/tasks'
```

Everything's installed:
```
$ bundle check
The Gemfile's dependencies are satisfied
```

The tasks are there:
```
$ rake -T
rake airbrake:deploy                      # Notify Airbrake of a new deploy
rake airbrake:install_heroku_deploy_hook  # Install a Heroku deploy hook to notify Airbrake of deploys
rake airbrake:test                        # Verify your gem installation by sending a test exception
```
...but calling the task fails due to a missing `Airbrake` constant:
```
$ rake airbrake:deploy
rake aborted!
NameError: uninitialized constant Airbrake
/usr/local/rvm/gems/ruby-2.4.1@airbroken/gems/airbrake-6.1.2/lib/airbrake/rake/tasks.rb:84:in `block (2 levels) in <top (required)>'
/usr/local/rvm/gems/ruby-2.4.1@global/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => airbrake:deploy
(See full trace by running task with --trace)
```

Now, the `airbrake:test` task also fails:
```
$ rake airbrake:test
rake aborted!
Don't know how to build task 'environment' (see --tasks)
/usr/local/rvm/gems/ruby-2.4.1@global/gems/rake-12.0.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => airbrake:test
(See full trace by running task with --trace)
```
...because of a dependency on a task called `environment`, which doesn't exist. So far, so fair.

However, even if such a task existed, and did what you might expect (load constants, set up configuration), I still don't think that the deploy task would work, because [that task looks like this (at the time of writing)](https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rake/tasks.rb#L63-L89):

```
...
63   desc 'Notify Airbrake of a new deploy'
64   task :deploy do
65     if defined?(Rails)
66       initializer = Rails.root.join('config', 'initializers', 'airbrake.rb')
67
68       # Avoid loading the environment to speed up the deploy task and try guess
69       # the initializer file location.
70       if initializer.exist? && Airbrake[:default].is_a?(Airbrake::NilNotifier)
71         load(initializer)
72       else
73         Rake::Task[:environment].invoke
74       end
75     end
76
77     deploy_params = {
78       environment: ENV['ENVIRONMENT'],
79       username: ENV['USERNAME'],
80       revision: ENV['REVISION'],
81       repository: ENV['REPOSITORY'],
82       version: ENV['VERSION']
83     }
84     promise = Airbrake.create_deploy(deploy_params)
85     promise.then do
86       puts "The #{deploy_params[:environment]} environment was deployed."
87     end
88     promise.rescue { |error| abort(error) }
89   end
 ...
```

Note that this task only runs the `environment` task *if* `Rails` is defined (and some other conditions).

I'm not sure what the correct solution is, but it feels like something along the lines of "load the environment task if `!defined?(Airbrake)`" ?

kyrylo:
Fixes #762 (`rake airbrake:deploy` doesn't seem to work in a non-Rails
project)

### DIFF
--- a/lib/airbrake/rake/tasks.rb
+++ b/lib/airbrake/rake/tasks.rb
@@ -1,73 +1,45 @@
+require 'airbrake-ruby'
+
 namespace :airbrake do
   desc 'Verify your gem installation by sending a test exception'
-  task test: :environment do
-    require 'pp'
-
-    begin
-      raise StandardError, 'Exception from the test Rake task'
-    rescue => ex
-      response = Airbrake.notify_sync(ex)
+  task test: (:environment if defined?(Rails)) do
+    unless Airbrake.configured?
+      raise Airbrake::Error, 'airbrake-ruby is not configured'
     end
 
-    notifiers = Airbrake.instance_variable_get(:@notifiers).map do |name, notif|
-      cfg = notif.instance_variable_get(:@config)
-      filters = notif.instance_variable_get(:@filter_chain)
-      "#{name}:\n  " + [cfg, filters].pretty_inspect
-    end.join("\n")
+    require 'pp'
 
-    if !response
-      puts <<-NORESPONSE.gsub(/^\s+\|/, '')
-        |Couldn't send a test exception. There are two reasons for this:
-        |
-        |1. Airbrake ignored this exception due to misconfigured filters
-        |2. Airbrake was configured to ignore the '#{Rails.env}' environment.
-        |   To fix this try one of the following:
-        |     * specify another environment via RAILS_ENV
-        |     * temporarily unignore the '#{Rails.env}' environment
-      NORESPONSE
-    elsif response['error']
+    response = Airbrake.notify_sync('Exception from the test Rake task')
+    if response['code']
       puts <<-ERROR.gsub(/^\s+\|/, '')
-        |Error occurred: #{response['error']}
-        |Make sure that your Project ID and Project Key are correct:
-        |https://github.com/airbrake/airbrake-ruby#project_id--project_key
+        |Couldn't send a test exception:
+        |#{response['type']}: #{response['message']} (#{response['code']})
+        |
+        |Possible problems:
+        |  1. Project ID or project key is incorrect
+        |  2. Exception was ignored due to misconfigured filters
+        |  3. The environment this task runs in is ignored by Airbrake
       ERROR
     else
       puts <<-OUTPUT.gsub(/^\s+\|/, '')
-        |[ruby]
-        |description: #{RUBY_DESCRIPTION}
-        |
-        |#{if defined?(Rails)
-             "[rails]\nversion: #{Rails::VERSION::STRING}"
-           elsif defined?(Sinatra)
-             "[sinatra]\nversion: #{Sinatra::VERSION}"
-           end}
-        |
-        |[airbrake]
-        |version: #{Airbrake::AIRBRAKE_VERSION}
-        |
-        |[airbrake-ruby]
-        |version: #{Airbrake::Notice::NOTIFIER[:version]}
-        |
-        |[notifiers]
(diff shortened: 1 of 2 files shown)


lib/airbrake/rake/tasks.rb
```diff
@@ -1,67 +1,42 @@
...
-    elsif response['error']
-      puts <<-ERROR.gsub(/^\s+\|/, '')
-        |Error occurred: #{response['error']}
-        |Make sure that your Project ID and Project Key are correct:
-        |https://github.com/airbrake/airbrake-ruby#project_id--project_key
-      ERROR
+    require 'pp'
+
+    response = Airbrake.notify_sync('Exception from the test Rake task')
+    raise response['error'] if response['error']
```
vmihailenco: AFAIK it is `response['message']` and you still need to check `response['code']` or `response['type']`

lib/airbrake/rake/tasks.rb
```diff
@@ -1,67 +1,42 @@
...
   desc 'Verify your gem installation by sending a test exception'
-  task test: :environment do
-    require 'pp'
-
-    begin
-      raise StandardError, 'Exception from the test Rake task'
-    rescue => ex
-      response = Airbrake.notify_sync(ex)
+  task test: (:environment if defined?(Rails)) do
+    if Airbrake[:default].is_a?(Airbrake::NilNotifier)
```
vmihailenco: Can you add a helper `if !Airbrake.configured` and use it everywhere?

lib/airbrake/rake/tasks.rb
```diff
@@ -1,67 +1,42 @@
...
+      raise 'The Airbrake library is loaded but not configured'
```
vmihailenco: Also it looks like Airbrake.notify could raise the same exception

lib/airbrake/rake/tasks.rb
kyrylo: https://github.com/airbrake/airbrake-ruby/pull/237

lib/airbrake/rake/tasks.rb
kyrylo: Refactored to a more general version.

lib/airbrake/rake/tasks.rb
kyrylo: I can't use `Airbrake.notify` because it's a task which can quit earlier than we can send an exception (if this is what you meant).

kyrylo:
PTAL

PR #780, opened 2017-08-25
fix reference to ActiveRecord

krisleech:
Sometimes I get the error `NameError:  uninitialized constant Airbrake::Rails::ActiveRecord::ConnectionAdapters`

I believe this is happening because `lib/airbrake/rails/active_record.rb` has been required  and thus the constant `Airbrake::Rails::ActiveRecord` is being referenced instead of `::ActiveRecord`

My guess is that the order the `initalizer`'s are run is not necessarily the order in which they are defined in the file and `initializer('airbrake.active_record')` which requires `airbrake/rails/active_record.rb` is being run before `initializer('airbrake.middleware')`.

By referencing back to the root scope, '::', it fixes the error.

* Rails 4.2
* Airbrake 6.2.1
* Airbrake-ruby 2.3.2

### DIFF
--- a/lib/airbrake/rails.rb
+++ b/lib/airbrake/rails.rb
@@ -19,11 +19,11 @@ module Airbrake
             ActionDispatch::DebugExceptions,
             Airbrake::Rack::Middleware
           )
-        elsif defined?(ActiveRecord)
+        elsif defined?(::ActiveRecord)
           # Insert after ConnectionManagement to avoid DB connection leakage:
           # https://github.com/airbrake/airbrake/pull/568
           app.config.middleware.insert_after(
-            ActiveRecord::ConnectionAdapters::ConnectionManagement,
+            ::ActiveRecord::ConnectionAdapters::ConnectionManagement,
             'Airbrake::Rack::Middleware'
           )
         else


kyrylo:
Interesting. Independently, I caught this error today! Looks like the build is failing, but it's not related to your PR. I will figure out how to fix it, then we would need to rebase this PR. I'll ping you.

kyrylo:
Apologies for the delay! I fixed the build. Could you please rebase on master? If this is green, I'll hit "merge" as soon as possible 👍

krisleech:
Rebased :+1:

krisleech:
Failed again: https://circleci.com/gh/airbrake/airbrake/1490#tests/containers/4

Looks like a Rubocop thing, which is odd since master passes.

kyrylo:
We always run latest Rubocop so apparently a new release was pushed with new cops. I'm looking into that.

kyrylo:
Should be fixed now. Could you rebase again? :)

krisleech:
done :D

kyrylo:
Woooo :)

PR #781, opened 2017-08-30
Rails5 - change position of Airbrake middleware

lowang:
move Airbrake to beginning of the stack - since in Rails 5, ActionDispatch::Executor
is handling db connections now, Airbrake needs to wrap it to catch ActiveRecord::ConnectionTimeoutError
also correct docs about Resque integration to allow both Airbrake & Resque to function properly (show errors in Resque, allow Resque retries)


kyrylo:
Hi @lowang! First of all, sorry for the delay in reviewing this. Secondly, there's a lot going on in this PR and unfortunately I will have to close it. However, either way I want to see this code merged. I would be happy to review a PR for Resque and Rails separately. Please let me know if you could do that. If not, don't worry, your contribution won't be lost.

airbrake.gemspec
```diff
@@ -1,4 +1,4 @@
-require './lib/airbrake/version'
+require File.expand_path("../lib/airbrake/version", __FILE__)
```
kyrylo: I'm not sure why this change is needed.

kyrylo:
1. I extracted a part of this PR to https://github.com/airbrake/airbrake/pull/788 but had to close it.
2. The readme example is simplified on purpose. We show how to use Airbrake with Resque and using `Multiple` isn't required.

PR #788, opened 2017-09-19
rails: insert our middleware before Executor on Rails 5

kyrylo:
Replaces https://github.com/airbrake/airbrake/pull/781

Description by @lowang:

> move Airbrake to beginning of the stack - since in Rails 5,
> ActionDispatch::Executor is handling db connections now, Airbrake
> needs to wrap it to catch ActiveRecord::ConnectionTimeoutError


kyrylo:
After some fiddling I realised that this change is breaking default behaviour (confirmed by failing tests: missing correct `context/url`). The way we currently insert middleware seems to be correct.

PR #799, opened 2017-09-29
Fix Sidekiq action(context)

thomey:
If ActiveJob is not used, you can for example have args that are `[1, 2, 3]`, which will not be caught by `return klass unless context[:job] && context[:job]['args'].first`.

This causes next line to fail with `no implicit conversion of String into Integer`.

Suggested fix: Only try to lookup 'job_class' if context[:job]['args'].first is a hash.


kyrylo:
Thanks!

kyrylo:
Released this fix in v7.0.2

Issue #604, opened 2016-09-20
user_attributes removed v5?


DiegoSalazar:
This was a valuable feature for us. Why was it removed?

kyrylo:
We send user information by default now.

> This was a valuable feature for us.

Could you elaborate?

DiegoSalazar:
> We send user information by default now.
Oh good to know!

It's helpful to know who was logged in when the error happened, we connect that information to other info we have of the user.

kyrylo:
Fantastic :)

Feel free to open new issues if you think we miss some information about users.

felixbuenemann:
@kyrylo It would be great if the v4 to v5 migration guide mentioned why it was removed.

kyrylo:
By now, I expect everyone to be already upgraded to v5. If you are still on v4, I wholeheartedly recommend to upgrade. We don't support v4 anymore (for almost a year already: https://github.com/airbrake/airbrake/issues/596). I'd be happy to accept a PR, if you don't mind, though.

felixbuenemann:
I could do a pr, if I knew what the new behavior is, which is still unclear to me.

Is airbrake now sending all attributes on the current_user model? What should I do to prevent that …

kyrylo:
[Like I previously said](https://github.com/airbrake/airbrake/issues/604#issuecomment-248542946), user information is sent [by default](https://github.com/airbrake/airbrake/blob/44d39c6bb33ce171e0d7633feb3e72acfef0b6a5/lib/airbrake/rack/user.rb#L41-L44).

felixbuenemann:
@kyrylo Thanks for pointing me at the right source code, I've prepared PR #805 to improve the docs.

PR #805, opened 2017-10-18
Document what happened to user_attributes in v5

felixbuenemann:
I was confused what happened to the `user_attributes` after upgrading from Airbrake v4 to v7 and the migration guide simply mentioned that the feature was removed, but didn't mention how user attributes were obtained in the new version.

I've updated the migration guide to document the new behavior as discussed with @kyrylo in #604.


felixbuenemann:
Test failures on CircleCI seem unrelated, looks like something is wrong with the rubocop setup.

kyrylo:
Thanks a bunch! As for the build failure, looks like a repo that we depended on got removed. I'll look into that, but I see no harm in merging this PR even if this build failed.

PR #810, opened 2017-12-08
Better handle detecting if ActiveRecord is actually loaded

urkle:
We have a product where we do not use ActiveRecord, however a gem is loading up active_record/errors.rb so it can handle errors for when a product does use AR.   This causes airbrake to incorrectly detect that activerecord is loaded and it blows up as the ConnectionAdapter code isn't actually loaded.  This extends the current check so that it better handles this scenario.

Related to #580

### DIFF
--- a/lib/airbrake/rails.rb
+++ b/lib/airbrake/rails.rb
@@ -19,7 +19,7 @@ module Airbrake
             ActionDispatch::DebugExceptions,
             Airbrake::Rack::Middleware
           )
-        elsif defined?(::ActiveRecord) && defined?(::ActiveRecord::ConnectionAdapters)
+        elsif defined?(::ActiveRecord::ConnectionAdapters::ConnectionManagement)
           # Insert after ConnectionManagement to avoid DB connection leakage:
           # https://github.com/airbrake/airbrake/pull/568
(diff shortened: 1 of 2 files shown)


lib/airbrake/rails.rb
```diff
@@ -19,7 +19,7 @@ class Railtie < ::Rails::Railtie
             ActionDispatch::DebugExceptions,
             Airbrake::Rack::Middleware
           )
-        elsif defined?(::ActiveRecord)
+        elsif defined?(::ActiveRecord) && defined?(::ActiveRecord::ConnectionAdapters)
```
kyrylo: Maybe we could get rid of `defined?(::ActiveRecord)` in favour of the new check? The new check covers the first one anyway.

urkle: hmm.. ahh didn't realize that..   maybe we should just be fully explicit then and do the whole class name.  Thus removing any possibility of it being incorrect.   I'll make that adjustment.

urkle: @kyrylo updated with a fixup.  And it looks like something is messed up with the circleci test rig as it's not installing certain gems

PR #812, opened 2017-12-12
spec: don't pass key for notice stubs

kyrylo:
In the most recent airbrake-ruby project key is passed through the Authorization
header, not a query parameter. [1]

[1]: https://github.com/airbrake/airbrake-ruby/commit/5d258a28993ca2b2649e02abd93dfa944f207063


kyrylo:
The build is expected to fail due to unknown circumstances which I'm trying to resolve in https://github.com/airbrake/airbrake/pull/811. However, it fixes a good chunk of failing tests related to the stubbing issue. Merging.

PR #817, opened 2018-01-04
Provide integration with sneakers gem

aiomaster:
Hi,
I want to add integration with the sneakers gem.
Sneakers uses bunny to connect to a rabbitmq server and provides workers.
Testsetup would be difficult, so I didn't made any tests.

### DIFF
--- a/airbrake.gemspec
+++ b/airbrake.gemspec
@@ -40,6 +40,7 @@ DESC
   s.add_development_dependency 'appraisal', '~> 2'
   s.add_development_dependency 'rack', '~> 1'
   s.add_development_dependency 'webmock', '~> 2'
+  s.add_development_dependency 'sneakers', '~> 2'
 
   # We still support Ruby 2.0.0+, but nokogiri 1.7.0+ doesn't.
   s.add_development_dependency 'nokogiri', '= 1.6.8.1'
--- a/lib/airbrake/sneakers.rb
+++ b/lib/airbrake/sneakers.rb
@@ -1,0 +1,18 @@
+module Airbrake
+  module Sneakers
+    ##
+    # Provides integration with Sneakers.
+    #
+    # @see https://github.com/jondot/sneakers
+    class ErrorReporter
+      def call(exception, worker = nil, **context)
+        Airbrake.notify(exception, context) do |notice|
+          notice[:context][:component] = 'sneakers'
+          notice[:context][:action] = worker.class.to_s
+        end
+      end
+    end
+  end
+end
+
+Sneakers.error_reporters << Airbrake::Sneakers::ErrorReporter.new
--- a/spec/unit/sneakers_spec.rb
+++ b/spec/unit/sneakers_spec.rb
@@ -1,0 +1,71 @@
+require 'spec_helper'
+require 'sneakers'
+require 'airbrake/sneakers'
+
+RSpec.describe Airbrake::Sneakers::ErrorReporter do
+  let(:worker) do
+    queue = instance_double('queue')
+    allow(queue).to receive_messages(
+      name: 'test-queue',
+      opts: {},
+      exchange: instance_double('exchange')
+    )
+
+    Class.new do
+      include Sneakers::Worker
+      from_queue 'defaults'
+
+      def work(_)
+        raise 'oops error'
+      end
+    end.new(queue)
+  end
+
+  let(:error) { StandardError.new('Something is wrong') }
+
+  let(:endpoint) { 'https://airbrake.io/api/v3/projects/113743/notices' }
+
+  def wait_for_a_request_with_body(body)
+    wait_for(a_request(:post, endpoint).with(body: body)).to have_been_made.once
+  end
+
+  before do
+    stub_request(:post, endpoint).to_return(status: 201, body: '{}')
+
+    Sneakers.configure(daemonize: true, log: '/dev/null')
+    Sneakers::Worker.configure_metrics
+  end
+
+  after do
+    Sneakers.clear!
+  end
+
+  it "should send a notice" do
+    handler = instance_double('handler')
+    expect(handler).to receive(:error).once
+    allow(worker.logger).to receive(:error)
+    worker.do_work(nil, nil, "msg", handler)
+
+    wait_for_a_request_with_body(/"message":"oops\serror"/)
+    wait_for_a_request_with_body(/test-queue/)
+    wait_for_a_request_with_body(/"component":"sneakers"/)
+  end
+
+  it "should support call without worker" do
+    Airbrake::Sneakers::ErrorReporter.new.call(error, message: 'my_glorious_message')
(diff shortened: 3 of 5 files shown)


kyrylo:
Thanks! Unit testing should be easy, see the shoryuken integration tests as a reference: https://github.com/airbrake/airbrake/blob/master/spec/unit/shoryuken_spec.rb

kyrylo:
Alternatively, Sidekiq tests https://github.com/airbrake/airbrake/blob/master/spec/unit/sidekiq_spec.rb

aiomaster:
It's not that easy I guess, but I'll try a little.

kyrylo:
Thank you. Please also rebase on master (I just fixed the build).

aiomaster:
While testing I've found, that the current sneakers version on github and the one on rubygems differ a little on the provided arguments to the error reporters.
I've updated the code, that it could work with both versions.

lib/airbrake/sneakers.rb
```diff
@@ -0,0 +1,18 @@
+module Airbrake
+  module Sneakers
+    ##
+    # Provides integration with Sneakers.
+    #
+    # @see https://github.com/jondot/sneakers
+    class ErrorReporter
+      def call(exception, worker = nil, **context)
```
kyrylo: Is double asterisk intentional?

airbrake.gemspec
```diff
@@ -40,6 +40,8 @@ DESC
   s.add_development_dependency 'appraisal', '~> 2'
   s.add_development_dependency 'rack', '~> 1'
   s.add_development_dependency 'webmock', '~> 2'
+  s.add_development_dependency 'sneakers', '~> 2'
+  s.add_development_dependency 'rr'
```
kyrylo: Why can't we use RSpec mocks (we use them for the rest of this project)?

aiomaster: Yes, we can. :)
I just copied some test setup from the sneakers gem and was lazy, but now I have rewritten it to use RSpec mocks like you suggested.

lib/airbrake/sneakers.rb
kyrylo: Isn't it for keyword arguments? Why can't we use a single asterisk?

spec/unit/sneakers_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+RSpec.describe Airbrake::Sneakers::ErrorReporter do
+  let(:worker) do
+    queue = instance_double('queue')
+    allow(queue).to receive_messages(
+      name: 'test-queue',
+      opts: {},
+      exchange: instance_double('exchange')
+    )
+
+    Sneakers.clear!
```
kyrylo: We should probably call this in an `after` block. Last executed test will leave Sneakers "dirty".

spec/unit/sneakers_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    Sneakers.configure(daemonize: true, log: '/dev/null')
+    Sneakers::Worker.configure_metrics
```
kyrylo: L15..L16 probably belong to a `before` block.

spec/unit/sneakers_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    Class.new do
+      include Sneakers::Worker
+      from_queue 'defaults'
+
+      def work(_)
+        raise 'oops error'
+      end
+    end.new(queue, Concurrent::ImmediateExecutor.new)
```
kyrylo: It seems like second parameter is optional. Can we omit it?

lib/airbrake/sneakers.rb
aiomaster: I simply want to support the call without a given worker as second argument:
```ruby
opts = { foo: 'bar' }
ErrorReporter.call(exception, opts)
```
so now with double asterisk `worker` is `nil` and `context` is `{ foo: 'bar' }`.
If I had just used a single asterisk on `context` the options would get assigned to `worker` and `context` is `nil`.

They added the `worker` parameter and I want to provide a solution that works with and without a given `worker` as second parameter.

This only works if the given options hash has symbolized keys, but this is the case:
https://github.com/jondot/sneakers/blob/7c7697757af239e2232e34c31bfd48e8bdacd596/lib/sneakers/worker.rb#L72

Another way would be to check the type of the worker argument, but I think it is more ugly.
```ruby
def call(exception, worker, context = nil)
  if worker.is_a?(Hash)
    context = worker
    worker = nil
  end
  # ...
end
```

spec/unit/sneakers_spec.rb
aiomaster: it seems we can

spec/unit/sneakers_spec.rb
aiomaster: valid point

spec/unit/sneakers_spec.rb
aiomaster: true

lib/airbrake/sneakers.rb
kyrylo: That makes sense. Can you please provide an example (just a link to the code) of a real job with keyword arguments?

kyrylo: By the way, I am fairly confident these details must be documented as a test. It's very tricky to understand so we must be sure that everything works as expected. Could you please a test(s) to make sure we support multiple arguments (different combinations).

aiomaster: The question is: Can or should we test multiple sneakers versions? It is easy to use the unreleased git version locally by adding `gem 'sneakers', github: 'jondot/sneakers'` to the Gemfile, but how could we manage that in a test?

I think I don't know, what you mean by "real job with keyword arguments". The keyword arguments are constructed here:
https://github.com/jondot/sneakers/blob/7c7697757af239e2232e34c31bfd48e8bdacd596/lib/sneakers/worker.rb#L72
and here the error_reporter is just called with it here:
https://github.com/jondot/sneakers/blob/7c7697757af239e2232e34c31bfd48e8bdacd596/lib/sneakers/error_reporter.rb#L18

But yeah, I can add a test, where the `Airbrake::Sneakers::ErrorReporter` is called directly with different arguments. Would that help?

aiomaster:
Ok, I've added two more tests to call the ErrorReporter directly with different arguments.
Thanks for your reviews and your feedback!

lib/airbrake/sneakers.rb
kyrylo: Testing this project is already hard enough. For simplicity we test only latest stable versions of libraries. So to answer your question, no we shouldn't.

Ok, I figured it out, let's stick to `**`.

kyrylo:
👍 👍 thanks for the new integration!

kyrylo:
Just tested on a Rails app, working good 👍

PR #828, opened 2018-03-01
LICENSE: update year from 2017 to 2018

kyrylo:


kyrylo:
The build was fixed via https://github.com/airbrake/airbrake/pull/829.

PR #831, opened 2018-04-12
Add a filter that ignore retryable Sidekiq jobs

### LINKED ISSUE #830: Ignoring errors from Sidekiq workers with upcoming retries
We noticed in our app that most of the Sidekiq jobs that failed would eventually succeed in one of the retries attempts that would happen, making the error report feel like noise rather than something we would have to fix right away.  Now we have [a custom filter class](https://gist.github.com/lucasmazza/31a0a3791f9a37da284344b6af3418cc) that would ignore errors from these jobs, leaving only errors from the final attempts to be collected by Airbrake. 

I think this could be useful for other users, and if it makes sense I can send a Pull Request adding the class and documenting how it can be enabled rather than leaving it on by default (which could be too big of a change for those who might not want this). What do you folks think? 😄

lucasmazza:
With this filter, errors from jobs tha will be retried will be ignored and only collected when Sidekiq exhausts its retries attempts

Closes #830

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,14 @@ require 'airbrake/sidekiq'
 If you required Sidekiq before Airbrake, then you don't even have to `require`
 anything manually and it should just work out-of-box.
 
+By default, Airbrake notifies of all errors, including reoccurring errors during a
+retry attempt. To filter out these errors and only get notified when Sidekiq has
+exhausted its retries you can add the `RetryableJobsFilter`:
+
+```ruby
+Airbrake.add_filter(Airbrake::Sidekiq::RetryableJobsFilter.new)
+```
+
 ### ActiveJob
 
 No additional configuration is needed. Simply ensure that you have configured
--- a/lib/airbrake/sidekiq/retryable_jobs_filter.rb
+++ b/lib/airbrake/sidekiq/retryable_jobs_filter.rb
@@ -1,0 +1,44 @@
+module Airbrake
+  module Sidekiq
+    ##
+    # Filter that can ignore notices from jobs that failed but will be retried
+    # by Sidekiq
+    class RetryableJobsFilter
+      if Gem::Version.new(::Sidekiq::VERSION) < Gem::Version.new('5.0.0')
+        require 'sidekiq/middleware/server/retry_jobs'
+        DEFAULT_MAX_RETRY_ATTEMPTS = \
+          ::Sidekiq::Middleware::Server::RetryJobs::DEFAULT_MAX_RETRY_ATTEMPTS
+      else
+        require 'sidekiq/job_retry'
+        DEFAULT_MAX_RETRY_ATTEMPTS = ::Sidekiq::JobRetry::DEFAULT_MAX_RETRY_ATTEMPTS
+      end
+
+      def call(notice)
+        job = notice[:params][:job]
+
+        notice.ignore! if retryable?(job)
+      end
+
+      private
+
+      def retryable?(job)
+        return false unless job && job['retry']
+
+        max_attempts = max_attempts_for(job)
+        job['retry_count'] < max_attempts
+      end
+
+      def max_attempts_for(job)
+        if job['retry'].is_a?(Integer)
+          job['retry']
+        else
+          max_retries
+        end
+      end
+
+      def max_retries
+        @max_retries ||= ::Sidekiq.options[:max_retries] || DEFAULT_MAX_RETRY_ATTEMPTS
+      end
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


lib/airbrake/sidekiq/retryable_jobs_filter.rb
```diff
@@ -0,0 +1,44 @@
+module Airbrake
+  module Sidekiq
+    ##
+    # Filter that can ignore notices from jobs that failed but will be retried
+    # by Sidekiq
+    class RetryableJobsFilter
+      if ::Sidekiq::VERSION < '5'
+        require 'sidekiq/middleware/server/retry_jobs'
+        DEFAULT_MAX_RETRY_ATTEMPTS = \
```
lucasmazza: I think this should be enough to support older Sidekiq versions (the constant was added somewhere around v2 but  moved places on v5) 🤔

lucasmazza:
The Circle Ci failure looks to be related to the `delayed_job_active_record` gem and the recent rails `5.2.0` release 😞

kyrylo:
Thanks for the PR! Could you please rebase on `master`? I *think* the issue should be gone.

lucasmazza:
@kyrylo rebased and pushed

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'spec_helper'
+
+if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.2.2')
+  require 'sidekiq'
+  require 'sidekiq/cli'
+  require 'airbrake/sidekiq'
+
+  RSpec.describe "airbrake/sidekiq/retryable_jobs_filter" do
+    subject(:filter) { Airbrake::Sidekiq::RetryableJobsFilter.new }
```
kyrylo: Why `subject` instead of `let`? We don’t seem to take the advantage of it anywhere, so `let` seems to be the right fit here.

README.md
```diff
@@ -321,6 +321,15 @@ require 'airbrake/sidekiq'
 If you required Sidekiq before Airbrake, then you don't even have to `require`
 anything manually and it should just work out-of-box.
 
+Airbrake will be notified of all errors from your Sidekiq workers, even when jobs
+will be retried in the future and might work as expected. To filter out these
+errors and only get notified when Sidekiq has exhausted its retries you can
+add the `RetryableJobsFilter` in your app:
```
kyrylo: What do you think of the following phrasing:

> By default, Airbrake notifies of all errors, including reoccurring errors during a retry attempt. To filter out these errors and only get notified when Sidekiq has exhausted its retries you can add the `RetryableJobsFilter`:

I think it’s a little shorter and conveys the same meaning.

lib/airbrake/sidekiq/retryable_jobs_filter.rb
```diff
@@ -0,0 +1,44 @@
+module Airbrake
+  module Sidekiq
+    ##
+    # Filter that can ignore notices from jobs that failed but will be retried
+    # by Sidekiq
+    class RetryableJobsFilter
+      if ::Sidekiq::VERSION < '5'
```
kyrylo: We should probably compare using `Gem::Version` as it is less error-prone.

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    def build_notice(job = nil)
+      Airbrake::Notice.new(Airbrake::Config.new, StandardError.new, job: job)
+    end
+
+    it "does not ignores notices that are not from jobs" do
```
kyrylo: Typo in `ignore`. The entire file has this typo.

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
lucasmazza: Semantics, I guess? The `filter` is still the system under test so `subject` seems better than `let` for me. 

I noticed now that most of the specs are using `subject { described_class.new }` and I can adopt that form if you think it's better

kyrylo: Unfortunately, RSpec doesn't provide any concrete pattern for this case. Same with BetterSpecs: http://www.betterspecs.org/#subject

They just say "you can use this" and that's it. Anyways, it seems to be a minor thing, so let's leave as is :)

kyrylo:
👍

PR #846, opened 2018-05-07
Fix Resque Failure Backend for non Active Job workers

Alfarlost:
In current implementation backend fails if first item in args is not a hash


kyrylo:
Sorry, I am not following. What's the scenario when the first item in args is not a hash? I tested with a dummy `CleanupJob` and passed a hash to it: `CleanupJob.perform_later(a: 1)`. The error was correctly reported to Airbrake with the following payload:

```
"args": [
    {
      "arguments": [
        {
          "_aj_symbol_keys": [
            "a"
          ],
          "a": "1"
        }
      ],
      "job_class": "CleanupJob",
      "job_id": "7215ddcc-28ab-437c-9ef7-13d09ac543db",
      "locale": "en",
      "queue_name": "default"
    }
  ],
```

I'm using Rails 4.2.10.

jeremy04:
A similar bug was fixed in the Sidekiq version: https://github.com/airbrake/airbrake/pull/799

Rails version: 5.1.4

Resque config:

```
require 'airbrake/resque'

Resque::Failure.backend = Resque::Failure::Airbrake

```

Test case: 
app/workers/dummy_job.rb:
```
class DummyJob
   @queue = :test
   def self.perform(arg1, arg2)
     raise "DummyJob Raised"
   end
end
```

`Resque.enqueue(DummyJob, *[1, 1])`

What we expect:
 an Airbrake message gets sent.

What happens: 
No airbrake message gets sent, silently fails.


When I enable the monkey patch, the Airbrake notification works.

kyrylo:
Thanks for the explanation. Makes sense.

PR #855, opened 2018-06-22
Do not track development env by default

lucianosousa:
It's just a suggestion.

Most of people/companies/projects I've worked always ignore dev env.

The reason is quite simple: None one is interested in see errors from someone else computer, eg: bad DB config, fix typo, and these sort of things


kyrylo:
I'd merge the change if it wasn't breaking the sign up flow. When a new Airbrake user signs up and installs a notifier, the first error they typically send is the test notice via:

```
bundle exec rake airbrake:test
```

This command runs in the development environment and will be ignored. This will create a lot of confusion so that's why we don't ignore it by default.

PR #856, opened 2018-07-05
[FEATURE] Add Ignorable Classes with retry attempts before ignore

dinks:
We had a specific use case (which we hope is somewhat common) where we wanted to ignore sending airbrakes for certain jobs until it tries a certain number of time. 

This PR (sidekiq filter) allows you do exactly this.

Hope it is useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,17 @@ has exhausted its retries you can add the `RetryableJobsFilter`:
 Airbrake.add_filter(Airbrake::Sidekiq::RetryableJobsFilter.new)
 ```
 
+To ignore certain classes `x` times use the `IgnorableClassFilter`
+
+```ruby
+Airbrake.add_filter(
+  Airbrake::Sidekiq::IgnorableClassFilter.new(
+    ignorable_classes: ['SomeIgnorableClass'],
+    retry_attempts_before_airbrake: x
+  )
+)
+```
+
 ### ActiveJob
 
 No additional configuration is needed. Simply ensure that you have configured
--- a/lib/airbrake/sidekiq.rb
+++ b/lib/airbrake/sidekiq.rb
@@ -1,5 +1,5 @@
 require 'airbrake/sidekiq/retryable_jobs_filter'
-require 'airbrake/sidekiq/ignorable_class_filter'
+require 'airbrake/sidekiq/ignorable_error_class_filter'
 
 module Airbrake
   module Sidekiq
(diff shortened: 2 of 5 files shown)


kyrylo:
Thanks for the PR!
Could you please rebase it on `master`? I fixed failures on JRuby.

I have a question about the purpose of this filter. It seems like it overlaps with `Airbrake::Sidekiq::RetryableJobsFilter`. Have you tried using it before writing your own filter? I wonder what's your use case?

README.md
```diff
@@ -329,6 +329,17 @@ has exhausted its retries you can add the `RetryableJobsFilter`:
 Airbrake.add_filter(Airbrake::Sidekiq::RetryableJobsFilter.new)
 ```
 
+To ignore certain classes `x` times use the `IgnorableClassFilter`
+
+```ruby
+Airbrake.add_filter(
+  Airbrake::Sidekiq::IgnorableClassFilter.new(
+    ignorable_classes: ['SomeIgnorableClass'],
```
kyrylo: Is the class here a worker name or an error class?

dinks: This is supposed to be an error class. Ill renamed this to be `ignorable_error_classes`

dinks:
`Airbrake::Sidekiq::RetryableJobsFilter` has the sole purpose of waiting until the max retries are done before sending to airbrake. (which might be waiting for weeks)

We had the problem that some error were raised due to connection errors or the resource being locked. This got fixed the second time around when the job retried. This PR lets you do exactly that.

It lets you define a set of error classes that could be ignored x times before sending an airbrake.

If you like to change the existing `RetryableJobsFilter`, I could try to do this.

README.md
kyrylo: What do you think about `ignorable_errors`? 

```
ignorable_errors: %w[ArgumentError RuntimeError]
```

It looks more terse and seems to convey the meaning, too.

kyrylo:
> If you like to change the existing RetryableJobsFilter, I could try to do this.

I can see that this PR has a lot of duplicate code with `RetryableJobsFilter`, so I think I would appreciate that. 

Based on your use case it seems like you want a more granular `RetryableJobsFilter`. For example, right now `RetryableJobsFilter` filters all error retries. If you give it a name of classes, it could only work for those classes.

```ruby
Airbrake::Sidekiq::RetryableJobsFilter.new(ignorable_errors: %w[ArgumentError RuntimeError])
```

dinks:
Well then this is a particular case of ours it seems. Ill close this PR for now. Thanks

kyrylo:
Just to clarify, I don't mind this functionality at all. I'm trying to say that embedding that logic into `RetryableJobsFilter` seems to be simpler than introducing a new filter. The feature is cool, but I think the implementation could be tweaked. Does it make sense?

PR #858, opened 2018-08-01
Added capistrano option that loads rails env when running rake

pnicholls:
## Problem 

Can't reference rails application classes from within the airbrake initialiser.

## Solution

Add `airbrake_load_environment` option that causes the `environment` option to be passed in when `rake` is run.

### DIFF
--- a/lib/airbrake/capistrano/capistrano2.rb
+++ b/lib/airbrake/capistrano/capistrano2.rb
@@ -18,7 +18,7 @@ module Airbrake
                 RACK_ENV=#{fetch(:rack_env, nil)} \
                 RAILS_ENV=#{fetch(:rails_env, nil)} \
 
-                bundle exec rake #{fetch(:airbrake_load_environment, nil)} airbrake:deploy \
+                bundle exec rake airbrake:deploy \
                   USERNAME=#{Shellwords.shellescape(ENV['USER'] || ENV['USERNAME'])} \
                   ENVIRONMENT=#{fetch(:airbrake_env, fetch(:rails_env, 'production'))} \
                   REVISION=#{current_revision.strip} \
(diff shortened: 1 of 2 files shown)


phumpal:
@pnicholls can you please correct `s/airbreak/airbrake/g` in both PR description and commit?

pnicholls:
Sorry folks - reverted Capistrano 2 implementation (https://github.com/airbrake/airbrake/pull/858/commits/88634e5349c5ece2d6fc6cfbebcc7d3d288fb64), that wasn't quite right. 

I have a project that has the following Airbrake initializer..

```ruby
  Airbrake.configure do |c|
    c.project_id = AppConstants.airbrake['project_id']
    c.project_key = AppConstants.airbrake['project_key']
    ....
```

Running `bundle exec rake airbrake:deploy` would fail as `AppConstants` would not have been loaded due to rake running without the `environment` option.

With `airbrake_load_env` set to `true` `airbrake:deploy` is now invoked with `bundle exec rake environment airbrake:deploy` which means `AppConstants` is loaded. 

I am not all that concerned if these PR gets approved or not, I figured the option might be helpful to others.

kyrylo:
Thanks, now it's more clear.

I think what you have here makes sense, however I'm a little wary of merging this because it's a special case. To aid that special case we introduce a global option, which sounds a bit like an overkill to me. No other behaviour will be attached to this option so it's a little too specific, it seems.

Loading environment is a slow and expensive process so I would try to find some other way to notify Airbrake of deploys. Have you tried requiring the file that defines `AppConstant` manually, from within your airbrake config file?

One other option is to define `AppConstants` in `config/initializers` and set the order you need with help of filename prefixes.

pnicholls:
> I think what you have here makes sense, however I'm a little wary of merging this because it's a special case. To aid that special case we introduce a global option, which sounds a bit like an overkill to me. No other behaviour will be attached to this option so it's a little too specific, it seems.

Yeah totally understand. 

> Loading environment is a slow and expensive process so I would try to find some other way to notify Airbrake of deploys. Have you tried requiring the file that defines AppConstant manually, from within your airbrake config file?

Won't work for this particular project but thanks for the idea non the less.

Cheers for the feedback, ill close this.

kyrylo:
Thanks for the PR, though. If anybody else stumbles upon this issue and wants something similar, please let me know.

PR #859, opened 2018-08-08
rails/active_job,rake: don't overwrite 'params'

kyrylo:
We shouldn't overwrite 'params' because it breaks filters, such as
`ContextFilter`, making `Airbrake.merge_context` unusable for these
integrations.


mmcdaris:
I checked for other instances of param assignment `notice[:params] = `, but these 2 look like the only ones.

Issue #883, opened 2018-12-15
Allow providing a custom class for Airbrake::Rack::ContextFilter to use when extracting user from request


pouellet:
It would be really useful to have an option to set the class that `Airbrake::Rack::ContextFilter` uses to try to determine which user is performing the request.

In one of our code base, we have some wrapper code surrounding the `current_user` method which may raise an exception under certain conditions. If this happens when the `Airbrake::Rack::User.extract` gets called, the Airbrake rack middleware ends up throwing an error and this results in a 500. 

While we can fix this by monkey patching `Airbrake::Rack::User.extract`, it would be cleaner to be able to specify which class to use to do the extraction. It would also allow the few folks that don't use OmniAuth or Devise to provide an implementation. Or to fix cases like this one #618.

I'm happy to do a PR if there's some interest in adding this.

kyrylo:
Sounds like a great idea. Feel free to submit a PR.

pouellet:
Thanks for the quick response, I'll give it a try.

pouellet:
@kyrylo I've got something going but I would like to get your input on two things:

1 - I would like to change the [`Airbrake::Rack::User.extract`](https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rack/user.rb#L10) method to return either a `Hash` or `nil` instead of a `User` object. The `Airbrake::Rack::ContextFilter` could then take the response directly [instead of having to call `to_json`](https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rack/context_filter.rb#L43) on the object returned. I think it would simplify the interface for people wanting to provide an implementation.

2 - I am not sure on where to allow users to specify the class to use. The only user configurable settings that I've found are all in `Airbrake::Config`, but that doesn't feel right to configure the middleware. Should I just allow configuring it directly on the `Airbrake::Rack::ContextFilter` class or would you rather have a more centralized configuration object?

kyrylo:
Resolved by https://github.com/airbrake/airbrake-ruby/pull/376 & https://github.com/airbrake/airbrake/pull/890.

PR #890, opened 2019-01-16
rack/middleware: move user extraction logic to its own filter

pouellet:
In order to give more control over how to find the current user and what
information to include in the payload, allow defining a class for the
rack middleware to use when extracting the current user payload from
the rack request.

### DIFF
--- a/spec/unit/rack/user_filter_spec.rb
+++ b/spec/unit/rack/user_filter_spec.rb
@@ -1,0 +1,36 @@
+require 'spec_helper'
+
+RSpec.describe Airbrake::Rack::UserFilter do
+  def env_for(url, opts = {})
+    Rack::MockRequest.env_for(url, opts)
+  end
+
+  subject { described_class.new }
+  let(:uri) { '/' }
+  let(:opts) { {} }
+
+  let(:notice) do
+    Airbrake.build_notice('oops').tap do |notice|
+      notice.stash[:rack_request] = Rack::Request.new(env_for(uri, opts))
+    end
+  end
+
+  let(:user_payload) { { username: 'bingo' } }
+  let(:user) { Airbrake::Rack::User.new(double(user_payload)) }
+
+  it "delegates extraction of the current user information" do
+    expect(Airbrake::Rack::User).to receive(:extract).and_return(user)
+    subject.call(notice)
+    expect(notice[:context][:user]).to eq(user_payload)
+  end
+
+  context "when no current user is found" do
+    let(:user) { Airbrake::Rack::User.new(double) }
+
+    it "does not include the user key in the payload" do
+      expect(Airbrake::Rack::User).to receive(:extract).and_return(user)
+      subject.call(notice)
+      expect(notice[:context].keys).not_to include(:user)
+    end
+  end
+end
--- a/spec/unit/rack/user_filter_spec.rb
+++ b/spec/unit/rack/user_filter_spec.rb
@@ -5,13 +5,9 @@ RSpec.describe Airbrake::Rack::UserFilter do
     Rack::MockRequest.env_for(url, opts)
   end
 
-  subject { described_class.new }
-  let(:uri) { '/' }
-  let(:opts) { {} }
-
   let(:notice) do
     Airbrake.build_notice('oops').tap do |notice|
-      notice.stash[:rack_request] = Rack::Request.new(env_for(uri, opts))
+      notice.stash[:rack_request] = Rack::Request.new(env_for('/', {}))
     end
   end
 
--- a/lib/airbrake/rack.rb
+++ b/lib/airbrake/rack.rb
@@ -1,4 +1,5 @@
 require 'airbrake/rack/user'
+require 'airbrake/rack/user_filter'
 require 'airbrake/rack/context_filter'
 require 'airbrake/rack/session_filter'
 require 'airbrake/rack/http_params_filter'
--- a/lib/airbrake/rack/context_filter.rb
+++ b/lib/airbrake/rack/context_filter.rb
@@ -35,13 +35,10 @@ module Airbrake
         add_framework_version(context)
 
         controller = request.env['action_controller.instance']
-        if controller
-          context[:component] = controller.controller_name
-          context[:action] = controller.action_name
-        end
+        return unless controller
 
-        user = Airbrake::Rack::User.extract(request.env)
-        notice[:context].merge!(user.as_json) if user
+        context[:component] = controller.controller_name
+        context[:action] = controller.action_name
       end
 
       private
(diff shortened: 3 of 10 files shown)


pouellet:
Example usage:

```ruby
class CustomAirbrakeUserExtractor
  class << self
    def extract(rack_env)
      user = current_user(rack_env)
      { id: user.id } if user
    end

    private

    def current_user(rack_env)
      controller.instance_variable_get(:@current_user)
    end
  end
end

Airbrake::Rack::ContextFilter.user_extractor_class = CustomAirbrakeUserExtractor
```

As discussed in #883, I'm not too happy with the `Airbrake::Rack::ContextFilter.user_extractor_class` part, but I'm not sure where would be the right location for that type of configuration.

kyrylo:
Thanks for the PR. I agree that `Airbrake::Rack::ContextFilter.user_extractor_class` doesn't look too good. I gave this a thought and came up with the following possible solution:

1. Make it possible to replace default filters (https://github.com/airbrake/airbrake-ruby/pull/375)
2. Extract the code that retrieves the user into a separate `UserFilter`:
        https://github.com/airbrake/airbrake/blob/12ac136938cfd9a00bca16bdbf1526b2e00fab8e/lib/airbrake/rack/context_filter.rb#L43-L44
3. Make `UserFilter` accept an "extractor" class: `UserFilter.new(CustomAirbrakeUserExtractor)`.

The benefit is that we utilise an already existing API, but `UserFilter` will have certain expectations (the extractor class must implement `extract`)

Does this make sense?

pouellet:
Pushing your idea further, would it make more sense to just add a call to allow removing filters instead of replacing the default filter on add? Someone out there might be adding two separate instances of the same filter class and the change in airbrake/airbrake-ruby#375 would break their workflow. So:

1. Make it possible to remove filters of a given class (or based on a condition defined in a Proc).
2. Extract the code that retrieves the user into a separate `UserFilter` (in order to be able to remove just this one instead of re-implementing the entire `Airbrake::Rack::ContextFilter`).

Then I could simply remove the default UserFilter and add my own CustomUserFilter to the filter chain.

What do you think?

kyrylo:
If we go with https://github.com/airbrake/airbrake-ruby/pull/375, then all you would need to do is to provide your `UserExtractor` class.

```ruby
# Inside the library we will do this:
Airbrake.add_filter(Airbrake::Filters::UserFilter.new(Airbrake::UserExtractor))

# In your code you will do this:
Airbrake.add_filter(Airbrake::Filters::UserFilter.new(UserExtractor))
```

If we go with your approach, we would need:

```ruby
# Inside airbrake
Airbrake.add_filter(Airbrake::Filters::UserFilter.new)

# In your code
Airbrake.delete_filter(Airbrake::Filters::UserFilter)
# or 
Airbrake.filters.delete_at(index)
```

> would it make more sense to just add a call to allow removing filters instead of replacing the default filter on add?

You're right, in theory it's a breaking change. That said, I took a look at our existing filters and I cannot come up with a reason why anyone would have two instances of it at the same time. I wouldn't encourage doing so in the user code due to possible side effects. We have the ["notice stash"](https://github.com/airbrake/airbrake-ruby#airbrakeadd_filter) API, so some filters sort of rely on each other (if you delete one, a group other filters will stop working).

If we are going to allow deleting default filters, then we would need to modify our public API:
* `Airbrake.delete_filter` needs to be added

In my opinion it's a big change. Up to now, if someone is unhappy with a default filter, and it's reasonable, we simply make it optional (not include it by default and suggest including it in the README). I'm a bit wary of doing it with `UserFilter` because it's fundamental and many people actually want it.

So after thing for a while about it, I'm trying to see how `Airbrake.delete_filter` would fit. I put up a PR https://github.com/airbrake/airbrake-ruby/pull/376

This solution isn't perfect. We can easily delete class-based filters but with Procs it's not trivial, since they have no names (you can technically pass Proc and it'll delete the first proc in the array but I'd rather not document this side effect). If you add multiple filters of the same class, only the first one will be deleted. I'm happy to live with this as long as we don't encourage this behaviour.

pouellet:
Thanks again for the quick feedback.

Either airbrake/airbrake-ruby#375 or airbrake/airbrake-ruby#376 will solve our problem and I am happy to adapt this PR to whichever of the two solutions you think is better.

I have updated the code to move the user extraction logic to a separate filter that takes as an argument an `extractor` class (this would align with airbrake/airbrake-ruby#375). If you prefer to go with airbrake/airbrake-ruby#376, I will simply remove that argument.

kyrylo:
Thinking more about it, I am not fond of "extractor" classes or other similar classes. The reason is that it's a new pattern that you have to learn, which adds to complexity. We already have filters and it would be nice to stick to them, so the API is simple and you don't need to have specific knowledge of filters (e.g. some filters require special arguments, some don't).

In theory `Airbrake.delete_filter` gives you more control about your data, so you don't send us something you're not willing to disclose. With https://github.com/airbrake/airbrake-ruby/pull/375 you are screwed.

Let's go with https://github.com/airbrake/airbrake-ruby/pull/376?

pouellet:
Sounds great to me.

I've squashed the commits and reverted most of the changes so that this PR is now only about moving the existing user extraction logic to its own filter.

spec/unit/rack/user_filter_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+RSpec.describe Airbrake::Rack::UserFilter do
+  def env_for(url, opts = {})
+    Rack::MockRequest.env_for(url, opts)
+  end
+
+  subject { described_class.new }
```
kyrylo: No need for this line, RSpec does this for us when we invoke `subject`.

kyrylo:
Nice work 👍 
I release airbrake-ruby [v3.1.0](https://github.com/airbrake/airbrake-ruby/tree/v3.1.0), which supports `Airbrake.delete_filter`. I'm going to release this gem now, too.

kyrylo:
Just released v8.0.1 with this change. Let me know if you encounter any troubles with this method that we built.

pouellet:
Thanks @kyrylo! I just had to figure out the right place to call the `delete_filter`, but once I got it right, it all works well. Since the Airbrake::Rack::Middleware gets initialized as part of the Rails app bootstrapping, it's necessary to remove the default filter in an `after_initialize` callback. The code looks something like this:

```ruby
class Application < Rails::Application
  config.after_initialize do
    # Replace default Airbrake User filter by our own implementation to prevent
    # Airbrake from calling the current_user method while reporting errors.
    # This needs to run after Rails initialize as the Airbrake::Rack::UserFilter
    # is added during Rails initialization.

    Airbrake.delete_filter(Airbrake::Rack::UserFilter)
    Airbrake.add_filter(AirbrakeCustomUserFilter.new)
  end
end
```

Thanks again for adding the `delete_filter` call to the API, this solution is a lot cleaner than the monkey patch we had in place to solve this. 👍

PR #892, opened 2019-01-29
rails: send SQL queries to Airbrake

kyrylo:


lib/airbrake/rails/active_record_subscriber.rb
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      def call(*args)
+        return unless (route = Thread.current[:airbrake_rails_route])
+
+        event = ActiveSupport::Notifications::Event.new(*args)
+        @notifier.notify_sql(
+          route: route,
+          method: Thread.current[:airbrake_rails_method],
+          sql: event.payload[:sql],
```
vmihailenco: Is this query normalized? e.g. `SELECT * FROM some WHERE id = ?` instead of `SELECT * FROM some WHERE id = 123`

kyrylo: Not normalised. Do we want to normalise it here or on the backend?

vmihailenco: If Rails provides normalized query then we should use it and send to the backend normalized query. If not we probably have to normalize queries on the server...

tagliala:
Hi,

commit ded45f3a9b69248913465c98b5417c6d0446537f is causing issues for my application which does not use ActiveRecord at all

```
~/.rvm/gems/ruby-2.6.1/gems/bootsnap-1.4.0/lib/bootsnap/load_path_cache/core_ext/active_support.rb:60:in `block in load_missing_constant': uninitialized constant Airbrake::Rack::Middleware::ActiveRecord (NameError)
```

Application: https://github.com/diowa/icare (downgraded to airbrake 7.4)

kyrylo:
Thanks for the report @tagliala. I fixed it in v8.1.1.

tagliala:
Thanks!

PR #914, opened 2019-02-27
rack/middleware: factor out hook installation code outside class

### LINKED ISSUE #909: NameError: uninitialized constant Airbrake::Rails
Hi,

I'm getting this error when trying to use Airbrake 8.1.4 in a project with ActiveRecord and Rack, but without Rails:

```
NameError: uninitialized constant Airbrake::Rails
  /Users/amikhailov/.rvm/rubies/ruby-2.5.3/lib/ruby/gems/2.5.0/gems/airbrake-8.1.4/lib/airbrake/rack/middleware.rb:136:in `install_active_record_hooks'
  /Users/amikhailov/.rvm/rubies/ruby-2.5.3/lib/ruby/gems/2.5.0/gems/airbrake-8.1.4/lib/airbrake/rack/middleware.rb:47:in `initialize'
```

kyrylo:
It was necessary to keep the hooks inside the middleware to support multiple
notifiers. Since we don't support this feautre anymore, we are feeling more
relaxed and can simplify the middleware. It was a horrible hack anyway, since we
had to have guards in `#initialize` and I'm glad it's going away.

`require` guarantees that we load the file only once, so it's safe to add
filters in the top level. We require these files from the railtie, and it also
inadvertently fixes #909 (NameError: uninitialized constant Airbrake::Rails).

### DIFF
--- a/lib/airbrake/rack/context_filter.rb
+++ b/lib/airbrake/rack/context_filter.rb
@@ -8,17 +8,6 @@ module Airbrake
       attr_reader :weight
 
       def initialize
-        @framework_version =
-          if defined?(::Rails) && ::Rails.respond_to?(:version)
-            { 'rails' => ::Rails.version }
-          elsif defined?(::Sinatra)
-            { 'sinatra' => Sinatra::VERSION }
-          else
-            {
-              'rack_version' => ::Rack.version,
-              'rack_release' => ::Rack.release
-            }
-          end
         @weight = 99
       end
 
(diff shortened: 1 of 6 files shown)


tagliala:
Hi again @kyrylo 😅 

Commit 4e5628f680aede2aeca5e422ccf54fca38fde8d8 broke again the compatibility with applications that do not use ActiveRecord

```
~/.rvm/gems/ruby-2.6.1/gems/bootsnap-1.4.1/lib/bootsnap/load_path_cache/core_ext/active_support.rb:76:in `block in load_missing_constant': uninitialized constant ActiveRecord (NameError)
```

I have temporarily downgraded to 8.1.4

Ref: https://github.com/airbrake/airbrake/pull/892#issuecomment-463267205

kyrylo:
Sorry about it, it's a known issue. Already fixed on `master` but not released yet.

tagliala:
Sorry, didn't check against the master branch 🤦‍♂️

Issue #922, opened 2019-03-05
Gem fails when not using ActiveRecord in Rails


lloydpick:
Hi,

In the newer releases (8.2.*) the code is making an assumption that ActiveRecord is available.

https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rails/active_record_subscriber.rb#L42

```
Traceback (most recent call last):
	23: from bin/rails:4:in `<main>'
	22: from bin/rails:4:in `require'
	21: from /Users/lloyd/.rvm/gems/ruby-2.6.0/gems/railties-4.2.11/lib/rails/commands.rb:17:in `<top (required)>'
	20: from /Users/lloyd/.rvm/gems/ruby-2.6.0/gems/railties-4.2.11/lib/rails/commands/commands_tasks.rb:39:in `run_command!'
	19: from /Users/lloyd/.rvm/gems/ruby-2.6.0/gems/railties-4.2.11/lib/rails/commands/commands_tasks.rb:67:in `console'
	18: from /Users/lloyd/.rvm/gems/ruby-2.6.0/gems/railties-4.2.11/lib/rails/commands/commands_tasks.rb:142:in `require_application_and_environment!'
	17: from /Users/lloyd/.rvm/gems/ruby-2.6.0/gems/railties-4.2.11/lib/rails/application.rb:328:in `require_environment!'
	16: from /Users/lloyd/code/something/config/environment.rb:5:in `<top (required)>'
	15: from /Users/lloyd/.rvm/gems/ruby-2.6.0/gems/railties-4.2.11/lib/rails/application.rb:352:in `initialize!'
	14: from /Users/lloyd/.rvm/gems/ruby-2.6.0/gems/railties-4.2.11/lib/rails/initializable.rb:54:in `run_initializers'
	13: from /Users/lloyd/.rvm/rubies/ruby-2.6.0/lib/ruby/2.6.0/tsort.rb:205:in `tsort_each'
	12: from /Users/lloyd/.rvm/rubies/ruby-2.6.0/lib/ruby/2.6.0/tsort.rb:226:in `tsort_each'
	11: from /Users/lloyd/.rvm/rubies/ruby-2.6.0/lib/ruby/2.6.0/tsort.rb:347:in `each_strongly_connected_component'
	10: from /Users/lloyd/.rvm/rubies/ruby-2.6.0/lib/ruby/2.6.0/tsort.rb:347:in `call'
[...]

PR #924, opened 2019-03-05
Reintroduce ActiveRecord guard clause

tlball:
If ActiveRecord has not been defined, loading ActiveRecordSubscriber will fail because of its call to ActiveRecord::Base.

Chose to add the guard clause to [lib/airbrake/rails.rb](https://github.com/airbrake/airbrake/blob/b21502812f7da58a8a3c508108c6a574cf1659f0/lib/airbrake/rails.rb) instead of to [airbrake/rails/active_record_subscriber](https://github.com/airbrake/airbrake/blob/b21502812f7da58a8a3c508108c6a574cf1659f0/lib/airbrake/rails/active_record_subscriber.rb), since that class name implies ActiveRecord is already available.

You can see a guard clause [used to exist](https://github.com/airbrake/airbrake/commit/4e5628f680aede2aeca5e422ccf54fca38fde8d8#diff-e09b896e9abbe09861d7846819a7bd56L32), but the code it referenced was removed and added into `Airbrake::Rails::ActiveRecordSubscriber` where it will break [on this line](https://github.com/airbrake/airbrake/blob/b21502812f7da58a8a3c508108c6a574cf1659f0/lib/airbrake/rails/active_record_subscriber.rb#L42) whenever that file tries to be loaded and ActiveRecord does not exist.

### DIFF
--- a/lib/airbrake/rails.rb
+++ b/lib/airbrake/rails.rb
@@ -7,7 +7,7 @@ module Airbrake
       initializer('airbrake.middleware') do |app|
         require 'airbrake/rails/action_controller_route_subscriber'
         require 'airbrake/rails/action_controller_notify_subscriber'
-        require 'airbrake/rails/active_record_subscriber'
+        require 'airbrake/rails/active_record_subscriber' if defined?(ActiveRecord)
 
         # Since Rails 3.2 the ActionDispatch::DebugExceptions middleware is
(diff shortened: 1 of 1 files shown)


kyrylo:
Thank you very much 🙂
Fixes https://github.com/airbrake/airbrake/issues/922.

PR #963, opened 2019-04-29
Rails 6: Pass middleware class instead of a string

codebycliff:
While attempting to upgrade to Rails `6.0.0.rc1`, I ran into an issue with Airbrake's middleware. It appears that [using strings or symbols for middleware class names is deprecated](https://github.com/rails/rails/commit/83b767ce). I have verified that this fixes the issue, though I'm not entirely sure of any side effects this may have within the Airbrake gem itself. Let me know if there is anything else I can do to help here. Here are the details:

* **Ruby**: 2.6.1
* **Rails**: 6.0.0.rc1
* **Stack trace**: 
image

### DIFF
--- a/lib/airbrake/rails.rb
+++ b/lib/airbrake/rails.rb
@@ -23,13 +23,13 @@ module Airbrake
           # https://github.com/airbrake/airbrake/pull/568
           app.config.middleware.insert_after(
             ::ActiveRecord::ConnectionAdapters::ConnectionManagement,
-            'Airbrake::Rack::Middleware'
+            Airbrake::Rack::Middleware
           )
         else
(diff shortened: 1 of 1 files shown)


kyrylo:
It's weird that we have this issue because not so long ago we had a fix for this: https://github.com/airbrake/airbrake/commit/eccf3c3409e8506fbf3e1eea5c3db55749d08167

Is there a chance that your app is on an old Airbrake version? v8.1.0 should already contain the fix: https://github.com/airbrake/airbrake/blob/master/CHANGELOG.md#v810-february-12-2019

codebycliff:
Yeah, that's the case. Sorry about that.

kyrylo:
Cool, let me know if there are any other issues.

PR #965, opened 2019-04-30
rails,rack: add the ability to send performance breakdown manually

kyrylo:
The new `Airbrake::Rack::Instrumentable` module adds
`airbrake_capture_timing`. This method expects the name of the method to
instrument. When such a method is called, Airbrake measures its timing and sends
the stats to Airbrake. The stats will be displayed in the performance breakdown
view of a route. Routes are determined automatically.

```ruby
class UsersController
  extend Airbrake::Rack::Instrumentable

  def index
    # ...
  end
  airbrake_capture_timing :index
end
```

Alternatively, you can measure performance within a method:

```ruby
def index
  Airbrake::Rack.capture_timing('operation name') do
    do_measured_stuff
  end

  do_other_stuff
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Airbrake Changelog
   ([#960](https://github.com/airbrake/airbrake/pull/960))
 * Rails: added support for Typhoeus for HTTP performance breakdown
   ([#961](https://github.com/airbrake/airbrake/pull/961))
+* Added `Airbrake::Rack.capture_timing` and `Airbrake::Rack::Instrumentable` for
+  manual performance measurements
+  ([#965](https://github.com/airbrake/airbrake/pull/965))
 
 ### [v9.1.0][v9.1.0] (April 17, 2019)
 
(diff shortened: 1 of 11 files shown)


vmihailenco:
Just FYI in gobrake & pybrake I call this spans, e.g. `start_span` and `end_span`. Overall you probably coud mimic opencensus API https://github.com/census-instrumentation/opencensus-ruby#adding-custom-trace-spans

```
OpenCensus::Trace.in_span "my_task" do |span|
  # Do stuff...
  
  OpenCensus::Trace.in_span "my_subtask" do |subspan|
    # Do other stuff
  end
end
```

Our API does not fully spans (especially hierarchy/ordering), but they are very close things.

kyrylo:
This is a higher level API. The low-level API calls them spans: https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/timed_trace.rb

This API also automatically captures routes, and I felt "span" would be too generic for this purpose.

PR #978, opened 2019-06-07
Fix broken retry count logic in Sidekiq RetryableJobsFilter

giddie:
It seems Sidekiq must have modified behaviour this filter relies on in recent versions. The existing code fails in two conditions:

1) When `job['retry_count']` is nil, which happens the first time a job fails. (There is no nil check.)
2) When `job['retry_count']` is not nil, the number represent how many retries have *already* been attempted. We must add 1 to this number, or `retry_count` will *always* be lower than `max_attempts`, even on the final retry when we want to actually notify.

### DIFF
--- a/lib/airbrake/sidekiq/retryable_jobs_filter.rb
+++ b/lib/airbrake/sidekiq/retryable_jobs_filter.rb
@@ -25,7 +25,8 @@ module Airbrake
         return false unless job && job['retry']
 
         max_attempts = max_attempts_for(job)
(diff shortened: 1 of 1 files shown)


kyrylo:
I can see the code here overlaps with https://github.com/airbrake/airbrake/pull/979. Seems like this wasn't intended. Could you clarify?

giddie:
Sorry, I forgot further commits on master would be added to the pull request. I'll start again.

giddie:
New PR at #980

PR #979, opened 2019-06-07
Add option to notify of sidekiq retries after a given number

giddie:
This is useful if we don't want to be spammed by a couple of retries,
but *do* want repeated warnings if a job continues to fail multiple
times.

### DIFF
--- a/lib/airbrake/sidekiq/retryable_jobs_filter.rb
+++ b/lib/airbrake/sidekiq/retryable_jobs_filter.rb
@@ -13,6 +13,10 @@ module Airbrake
         DEFAULT_MAX_RETRY_ATTEMPTS = ::Sidekiq::JobRetry::DEFAULT_MAX_RETRY_ATTEMPTS
       end
 
+      def initialize(max_retries: nil)
+        @max_retries = max_retries
+      end
+
       def call(notice)
         job = notice[:params][:job]
 
@@ -30,7 +34,9 @@ module Airbrake
       end
 
       def max_attempts_for(job)
-        if job['retry'].is_a?(Integer)
+        if @max_retries
+          @max_retries
+        elsif job['retry'].is_a?(Integer)
           job['retry']
         else
           max_retries
--- a/spec/unit/sidekiq/retryable_jobs_filter_spec.rb
+++ b/spec/unit/sidekiq/retryable_jobs_filter_spec.rb
@@ -3,8 +3,9 @@ if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.2.2')
   require 'sidekiq/cli'
   require 'airbrake/sidekiq'
 
-  RSpec.describe "airbrake/sidekiq/retryable_jobs_filter" do
-    subject(:filter) { Airbrake::Sidekiq::RetryableJobsFilter.new }
+  RSpec.describe Airbrake::Sidekiq::RetryableJobsFilter do
+    subject(:filter) { described_class.new }
+
     def build_notice(job = nil)
       Airbrake::Notice.new(StandardError.new, job: job)
     end
--- a/spec/unit/sidekiq/retryable_jobs_filter_spec.rb
+++ b/spec/unit/sidekiq/retryable_jobs_filter_spec.rb
@@ -13,13 +13,13 @@ if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.2.2')
     it "does not ignore notices that are not from jobs" do
       notice = build_notice
       filter.call(notice)
-      expect(build_notice).to_not be_ignored
+      expect(notice).to_not be_ignored
     end
 
     it "does not ignore notices from jobs that have retries disabled" do
       notice = build_notice('retry' => false)
       filter.call(notice)
-      expect(build_notice).to_not be_ignored
+      expect(notice).to_not be_ignored
     end
 
     it "ignore notices from jobs that will be retried" do
--- a/spec/unit/sidekiq/retryable_jobs_filter_spec.rb
+++ b/spec/unit/sidekiq/retryable_jobs_filter_spec.rb
@@ -38,5 +38,27 @@ if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.2.2')
       notice = build_notice('retry' => 3)
       expect { filter.call(notice) }.to_not raise_error
     end
+
+    context 'with max_retries = 2' do
+      subject(:filter) { described_class.new(max_retries: 2) }
+
+      it "ignores notices when retry_count is null" do
+        notice = build_notice('retry' => 5)
+        filter.call(notice)
+        expect(notice).to be_ignored
+      end
+
+      it "ignores notices when retry_count is 0" do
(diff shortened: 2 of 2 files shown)


lib/airbrake/sidekiq/retryable_jobs_filter.rb
```diff
@@ -13,6 +13,10 @@ class RetryableJobsFilter
         DEFAULT_MAX_RETRY_ATTEMPTS = ::Sidekiq::JobRetry::DEFAULT_MAX_RETRY_ATTEMPTS
       end
 
+      def initialize(notify_after: nil)
```
kyrylo: It's hard to guess that this option expects a number. It looks more like a callback to me, so I would think it wants a Proc. The option is also related to retries but nothing in the name suggests that. I believe we need to find a better fit.

Maybe it could be: `max_retries`? We would set it to the `max_retries` method call by default. Then the `max_attempts_for` would use `@max_retries` instead of `max_retries`. Does that make sense?

P.S. We need unit tests here as well.

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
```diff
@@ -32,5 +32,19 @@ def build_notice(job = nil)
       filter.call(notice)
       expect(build_notice).to_not be_ignored
     end
+
+    context 'with max_retries specified' do
+      it "ignores notices below max_retries" do
+        notice = build_notice('retry' => 5, 'retry_count' => 1, 'max_retries' => 2)
+        filter.call(notice)
+        expect(build_notice).to_not be_ignored
```
kyrylo: According to the spec description this has to be `expect(build_notice).to be_ignored` (unless I am missing something).

lib/airbrake/sidekiq/retryable_jobs_filter.rb
```diff
@@ -29,7 +33,9 @@ def retryable?(job)
       end
 
       def max_attempts_for(job)
-        if job['retry'].is_a?(Integer)
+        if @max_retries
```
kyrylo: Can this actually ever be `true`? Probably only when we initialise like this: `RetryableJobsFilter.new(max_retries: nil)`. That will look a bit confusing IMO.

In all other cases we will have the value defaulted to `DEFAULT_MAX_RETRY_ATTEMPTS`.

lib/airbrake/sidekiq/retryable_jobs_filter.rb
```diff
@@ -13,6 +13,10 @@ class RetryableJobsFilter
         DEFAULT_MAX_RETRY_ATTEMPTS = ::Sidekiq::JobRetry::DEFAULT_MAX_RETRY_ATTEMPTS
       end
 
+      def initialize(max_retries: nil)
```
kyrylo: Perhaps, it would be better not to shadow the method call to avoid headaches with precedence. I think this might work:

```rb
def initialize(max_retries: nil)
  @max_retries = max_retries || find_max_retries
end
```

Where `find_max_retries` is the current `max_retries` method. This will fix the condition on line 36.

lib/airbrake/sidekiq/retryable_jobs_filter.rb
giddie: Well, `@max_retries` will be nil by default if `max_retries` is not passed when creating the object, because it's specified in the parameter list of `initialize`.

lib/airbrake/sidekiq/retryable_jobs_filter.rb
giddie: Mmm the problem is that the precedence must go: `@max_retries` (the option) => `job['retry']` => `max_retries` (the method). I don't think we can combine the `@max_retries` option with the value returned by `max_retries` (or `find_max_retries`) and still retain that order of precedence.

kyrylo:
Could you rebase? There's currently a conflict.

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
```diff
@@ -30,7 +30,23 @@ def build_notice(job = nil)
     it "does not ignore notices from jobs that will not be retried" do
       notice = build_notice('retry' => 5, 'retry_count' => 5)
       filter.call(notice)
-      expect(build_notice).to_not be_ignored
+      expect(notice).to_not be_ignored
+    end
+
+    context 'with max_retries specified' do
+      subject(:filter) { Airbrake::Sidekiq::RetryableJobsFilter.new(max_retries: 2) }
```
kyrylo:
```suggestion
      subject(:filter) { described_class.new(max_retries: 2) }
```

kyrylo:
Could you help me understand the status of this PR? Do you need any help?

giddie:
Sorry, my bad. I had missed that the specs were failing.

giddie:
Mmm; looks like that ruby-2.2 CI failure is not related to this branch.

kyrylo:
Yeah, looks like i18n decided to drop support of Rubies below 2.3:
https://github.com/ruby-i18n/i18n/commit/213641fe9a0c8f59c5b7d57b2e437d9cf4cfc825

I'll resolve that separately and get back to your PR.

kyrylo:
Should be fixed on master now. Could you please rebase? Sorry for the trouble.

giddie:
No problem :)

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
kyrylo: Oh, I just noticed this wasn't actually fixed. Should still be `expect(notice).to be_ignored` according to the spec description. P.S. Also note that it's `build_notice` instead of `notice`. We basically introduce the same bug as on line 22 where you fixed it 😉

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
```diff
@@ -37,5 +38,27 @@ def build_notice(job = nil)
...
+        expect(notice).to be_ignored
+      end
+
+      it "ignores notices when retry_count is 0" do
+        notice = build_notice('retry' => 5, 'retry_count' => 0)
+        filter.call(notice)
+        expect(build_notice).to_not be_ignored
+      end
+
+      it "does not ignore notices is 1" do
```
kyrylo: This probably means `does not ignore notices when retry_count is 1`.

spec/unit/sidekiq/retryable_jobs_filter_spec.rb
```diff
@@ -37,5 +38,27 @@ def build_notice(job = nil)
...
+        notice = build_notice('retry' => 5, 'retry_count' => 1)
+        filter.call(notice)
+        expect(build_notice).to_not be_ignored
```
kyrylo:
```suggestion
        expect(notice).to_not be_ignored
```

kyrylo:
This comment is hidden for some reason: https://github.com/airbrake/airbrake/pull/979#discussion_r298067888

Highlighting it here since it's a bit hard to spot.

giddie:
Oh sorry, I think I lost track of those when I was rebasing.

PR #1007, opened 2019-09-06
Fix for #try_current_user when the Rails controller method throws an …

pbzymek:
Hey, this solves an issue I run into when the controller `current_user` method throws an exception.

### DIFF
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -27,6 +27,8 @@ module Airbrake
         return unless controller.respond_to?(:current_user, true)
         return unless [-1, 0].include?(controller.method(:current_user).arity)
         controller.__send__(:current_user)
+      rescue Exception => e
+        nil
       end
       private_class_method :try_current_user
 
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -22,16 +22,17 @@ module Airbrake
         new(user) if user
       end
 
-      # rubocop:disable Lint/RescueException
       def self.try_current_user(rack_env)
         controller = rack_env['action_controller.instance']
         return unless controller.respond_to?(:current_user, true)
         return unless [-1, 0].include?(controller.method(:current_user).arity)
-        controller.__send__(:current_user)
-      rescue Exception => _e
-        nil
+        begin
+          controller.__send__(:current_user)
+        rescue Exception => _e # rubocop:disable Lint/RescueException
+          nil
+        end
       end
-      # rubocop:enable Lint/RescueException
+
       private_class_method :try_current_user
 
       def initialize(user)
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -22,6 +22,7 @@ module Airbrake
         new(user) if user
       end
 
+      # rubocop:disable Lint/RescueException
       def self.try_current_user(rack_env)
         controller = rack_env['action_controller.instance']
         return unless controller.respond_to?(:current_user, true)
@@ -30,6 +31,7 @@ module Airbrake
       rescue Exception => e
         nil
       end
+      # rubocop:enable Lint/RescueException
       private_class_method :try_current_user
 
       def initialize(user)
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -32,7 +32,6 @@ module Airbrake
           nil
         end
       end
-
       private_class_method :try_current_user
 
       def initialize(user)
--- a/lib/airbrake/rack/user.rb
+++ b/lib/airbrake/rack/user.rb
@@ -28,7 +28,7 @@ module Airbrake
         return unless controller.respond_to?(:current_user, true)
         return unless [-1, 0].include?(controller.method(:current_user).arity)
         controller.__send__(:current_user)
-      rescue Exception => e
+      rescue Exception => _e
         nil
       end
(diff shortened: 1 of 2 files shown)


kyrylo:
Why would we want to silence exceptions in your method? I don't think it's a good idea to do that by default, but perhaps I am missing something. If your `current_user` throws an exception, shouldn't you fix it instead?

pbzymek:
this happens when ContextFilter is being registered in Rack, the side effect is that Airbrake gem doesn't work at all, even when authentication is not required for a given controller method.

Regarding my `current_user` it throws the exception by design and should be called in certain controller actions. When `current_user` is called call out of scope in rack middleware I'd be great to make it work that Airbrake registers with exception silencing. Let me know if this makes sens for you.

kyrylo:
Sorry, I don't think I'm following. Do you have an example for this:

>  even when authentication is not required for a given controller method.

Are you referring to https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rack/context_filter.rb ? I'm not sure how that file is related to anything being discussed here. If it's not included, your error report will simply lack some of the information, but the library should work just fine.

> Regarding my current_user it throws the exception by design and should be called in certain controller actions.

Making `current_user` raise in certain situation is not something I would expect as a Rails developer. You often use this call for conditional logic in the views: `if current_user then render(FOO) else render(BAR)`.

pbzymek:
> Are you referring to https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rack/context_filter.rb ? I'm not sure how that file is related to anything being discussed here. If it's not included, your error report will simply lack some of the information, but the library should work just fine.

This is about `try_current_user`(https://github.com/airbrake/airbrake/blob/89cb00247553764219835e6b09ac24de3c96a86b/lib/airbrake/rack/user_filter.rb#L18 https://github.com/airbrake/airbrake/blob/89cb00247553764219835e6b09ac24de3c96a86b/lib/airbrake/rack/user.rb#L21 ) method that is being called when `airbrake/airbrake` gem registers in Rack. In my case the `current_user` is being called before any controller method and as a result I'm not getting exception in my error notification system. Simply getting an unhandled exception at Rack level disables `airbrake/airbrake` gem. 

I do understand that normally `current_user` shouldn't throw exceptions. However, if this brekes the gem functionality I guess it's worth handling.

pbzymek:
Hey, any chance we could follow up on this?

lib/airbrake/rack/user.rb
```diff
@@ -22,12 +22,16 @@ def self.extract(rack_env)
...
+      # rubocop:disable Lint/RescueException
       def self.try_current_user(rack_env)
         controller = rack_env['action_controller.instance']
         return unless controller.respond_to?(:current_user, true)
         return unless [-1, 0].include?(controller.method(:current_user).arity)
         controller.__send__(:current_user)
+      rescue Exception => _e
+        nil
       end
+      # rubocop:enable Lint/RescueException
```
kyrylo: We can move this to line 31

```rb
rescue Exception => _e # rubocop:enable Lint/RescueException
```

spec/unit/rack/user_spec.rb
```diff
@@ -60,6 +60,15 @@ def env_for(url, opts = {})
         end
       end
 
+      context 'and it throw exception' do
```
kyrylo:
```suggestion
      context 'and when it throws an exception' do
```

spec/unit/rack/user_spec.rb
```diff
@@ -60,6 +60,15 @@ def env_for(url, opts = {})
...
+        it 'returns nil' do
+          allow(controller).to receive(:current_user) { raise 'Error' }
```
kyrylo: No need for `'Error'`. Just `raise` is enough.

lib/airbrake/rack/user.rb
```diff
@@ -22,12 +22,16 @@ def self.extract(rack_env)
         new(user) if user
       end
 
+      # rubocop:disable Lint/RescueException
       def self.try_current_user(rack_env)
         controller = rack_env['action_controller.instance']
         return unless controller.respond_to?(:current_user, true)
         return unless [-1, 0].include?(controller.method(:current_user).arity)
         controller.__send__(:current_user)
+      rescue Exception => _e
```
kyrylo: We probably want to wrap line 30 only, not the whole method. This way we don't accidentally catch something from our own code.

spec/unit/rack/user_spec.rb
kyrylo: Actually, I think we can even use this:

```
allow(controller).to receive(:current_user).and_raise
```

lib/airbrake/rack/user.rb
```diff
@@ -26,8 +26,13 @@ def self.try_current_user(rack_env)
...
         return unless controller.respond_to?(:current_user, true)
         return unless [-1, 0].include?(controller.method(:current_user).arity)
-        controller.__send__(:current_user)
+        begin
+          controller.__send__(:current_user)
+        rescue Exception => _e # rubocop:disable Lint/RescueException
+          nil
+        end
       end
+
```
kyrylo: Sorry, I don't want to sound nitpicky but let's not add this line. I've been using this style consistently across the codebase, so that it's easy to see that the private method is just above.

pbzymek: sure, no problem

kyrylo:
Thank you, I'll release this sometime during this week.

PR #1065, opened 2020-01-23
Fix keyword argument error in Sneakers integration

BrianHawley:
The Sneakers::ErrorReporter worker_error method context_hash parameter
is Hash data, not keyword parameters. Declare our handler accordingly.

### DIFF
--- a/lib/airbrake/sneakers.rb
+++ b/lib/airbrake/sneakers.rb
@@ -10,12 +10,15 @@ module Airbrake
       # @see https://github.com/airbrake/airbrake/issues/850
       IGNORED_KEYS = %i[delivery_tag consumer channel].freeze
 
-      def call(exception, worker = nil, **context)
+      # rubocop:disable Style/OptionalArguments
+      def call(exception, worker = nil, context)
+        # Later versions add a middle argument.
         Airbrake.notify(exception, filter_context(context)) do |notice|
           notice[:context][:component] = 'sneakers'
           notice[:context][:action] = worker.class.to_s
         end
       end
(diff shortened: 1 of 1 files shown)


BrianHawley:
This fixes Ruby 2.7 keyword argument errors.

lib/airbrake/sneakers.rb
```diff
@@ -10,12 +10,15 @@ class ErrorReporter
       # @see https://github.com/airbrake/airbrake/issues/850
       IGNORED_KEYS = %i[delivery_tag consumer channel].freeze
 
-      def call(exception, worker = nil, **context)
+      # rubocop:disable Style/OptionalArguments
+      def call(exception, worker = nil, context)
+        # Later versions add a middle argument.
```
kyrylo: Could you elaborate on this? I'm not a Sneakers user myself and this integration was contributed by someone else.
I can see that Sneakers actually always passes 2nd argument:
https://github.com/jondot/sneakers/blob/7c059d1490e4dde43f98b801c38e4e0df7ac458c/lib/sneakers/error_reporter.rb#L18

Perhaps, we should just delete `= nil` so that the signature is:

`def call(exception, worker, context)`

BrianHawley: That won't work with some versions of sneakers. I'm not a sneakers user myself but I tracked down the code that calls this handler and reviewed its commit history. At one point they added a second argument, which they always pass in versions since then. For versions before then they only pass two arguments, which for this handler correspond to the first and last argument.

kyrylo: Thanks, that makes sense.

PR #1064, opened 2020-01-23
Test against ruby-2.7.0 final

BrianHawley:
Also fixed a yaml typo.

PR #1066, opened 2020-01-23
Enhancements to airbrake_capture_timing

### LINKED ISSUE #1062: Airbrake::Rack::Instrumentable airbrake_capture_timing limits
The `Airbrake::Rack::Instrumentable#airbrake_capture_timing` method has some limits:

It doesn't work with methods ending with "!", "?", or "=", the way `alias_method_chain` does.

It has keyword argument warnings when wrapping a method with such on Ruby 2.7+.

Because it's done in method chain style, it will stack overflow when applied to prepended methods, as you see in Rails 5+. On the other hand, if it were implemented with `Module#prepend`, it would stack overflow with chained methods, as you see in Rails < 5. We need to support both techniques, with an option and a settable default for that option so people can choose the technique that is compatible with the majority of their codebase.

BrianHawley:
- Now supports wrapping the target method with alias_method_chain style
  or Module prepend style. Which one you should use depends on whether
  the method you're wrapping is chained or prepended already.
- Prevents you from wrapping a method more than once in chain mode,
  because that leads to stack overflows.
- Can handle method names ending in "?", "!", or "=".
- Can wrap operator methods (in prepend mode, at least).
- Uses Ruby 2.7+ style method forwarding when that's supported.
- Defines the wrappers using module_eval so we don't have closure
  references left in memory.

[Fixes #1062]

### DIFF
--- a/lib/airbrake/rack/instrumentable.rb
+++ b/lib/airbrake/rack/instrumentable.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Airbrake
   module Rack
     # Instrumentable holds methods that simplify instrumenting Rack apps.
@@ -15,12 +17,118 @@ module Airbrake
     # @since v9.2.0
     module Instrumentable
       def airbrake_capture_timing(method_name, label: method_name.to_s)
-        alias_method "#{method_name}_without_airbrake", method_name
+        instrumentable = ::Airbrake::Rack::Instrumentable
+        if instrumentable.should_prepend?(self, method_name)
+          instrumentable.prepend_capture_timing(self, method_name, label)
+        else
+          instrumentable.chain_capture_timing(self, method_name, label)
+        end
+        method_name
+      end
 
-        define_method(method_name) do |*args|
-          Airbrake::Rack.capture_timing(label) do
-            __send__("#{method_name}_without_airbrake", *args)
+      # @api private
+      def __airbrake_capture_timing_module__
+        # Module used to store prepended wrapper methods, saved as an instance
+        # variable so each target class/module gets its own module. This just
+        # a convenience to avoid prepending a lot of anonymous modules.
+        @__airbrake_capture_timing_module__ ||= ::Module.new
+      end
+      private :__airbrake_capture_timing_module__
+
+      # Using api private self methods so they don't get defined in the target
+      # class or module, but can still be called by the above method.
+
+      # @api private
+      def self.should_prepend?(klass, method_name)
+        # Don't chain already-prepended or operator methods.
+        klass.module_exec do
+          self_class_idx = ancestors.index(self)
+          method_owner_idx = ancestors.index(instance_method(method_name).owner)
+          method_owner_idx < self_class_idx || !(/\A\W/ =~ method_name).nil?
+        end
+      end
+
+      # @api private
+      def self.prepend_capture_timing(klass, method_name, label)
+        args = method_signature
+        visibility = method_visibility(klass, method_name)
+
+        # Generate the wrapper method.
+        klass.module_exec do
+          mod = __airbrake_capture_timing_module__
+          mod.module_exec do
+            module_eval <<-RUBY, __FILE__, __LINE__ + 1
+              def #{method_name}(#{args})
+                Airbrake::Rack.capture_timing(#{label.to_s.inspect}) do
+                  super
+                end
+              end
+              #{visibility} :#{method_name}
+            RUBY
           end
+          prepend mod
+        end
+      end
+
+      # @api private
+      def self.chain_capture_timing(klass, method_name, label)
+        args = method_signature
+        visibility = method_visibility(klass, method_name)
+
+        # Generate the wrapper method.
+        aliased = method_name.to_s.sub(/([?!=])$/, '')
+        punctuation = Regexp.last_match(1)
+        wrapped_method_name = "#{aliased}_without_airbrake#{punctuation}"
+        needs_removal = method_needs_removal(klass, method_name)
+        klass.module_exec do
+          alias_method wrapped_method_name, method_name
+          remove_method method_name if needs_removal
+          module_eval <<-RUBY, __FILE__, __LINE__ + 1
+            def #{method_name}(#{args})
+              Airbrake::Rack.capture_timing(#{label.to_s.inspect}) do
+                __send__("#{aliased}_without_airbrake#{punctuation}", #{args})
+              end
+            end
+            #{visibility} :#{method_name}
+          RUBY
+        end
+      end
+
+      # @api private
+      def self.method_visibility(klass, method_name)
+        klass.module_exec do
+          if protected_method_defined?(method_name)
+            "protected".freeze
+          elsif private_method_defined?(method_name)
+            "private".freeze
+          else
+            "public".freeze
+          end
+        end
+      end
+
+      # @api private
+      # A method instead of a constant so it isn't accessible in the target.
+      if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.7")
+        def self.method_signature
+          "*args, **kw_args, &block".freeze
+        end
+      else
+        def self.method_signature
+          "*args, &block".freeze
+        end
+      end
+
+      # @api private
+      if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.6")
+        def self.method_needs_removal(klass, method_name)
+          klass.method_defined?(method_name, false) ||
+            klass.private_method_defined?(method_name, false)
+        end
+      else
+        def self.method_needs_removal(klass, method_name)
+          klass.instance_methods(false).include?(method_name) ||
+            klass.private_instance_methods(false).include?(method_name)
         end
       end
     end
--- a/spec/unit/rack/instrumentable_spec.rb
+++ b/spec/unit/rack/instrumentable_spec.rb
@@ -11,6 +11,8 @@ RSpec.describe Airbrake::Rack::Instrumentable do
         def method; end
         airbrake_capture_timing :method
 
+        def foo; end
+
         def method_with_arg(a); end
         airbrake_capture_timing :method_with_arg
 
@@ -25,7 +27,128 @@ RSpec.describe Airbrake::Rack::Instrumentable do
 
         def method_with_kwargs(foo:, bar:); end
         airbrake_capture_timing :method_with_kwargs
+
+        def method_with!(val); end
+        airbrake_capture_timing :method_with!
+
+        def method_with?(val); end
+        airbrake_capture_timing :method_with?
+
+        attr_writer :method_with
+        airbrake_capture_timing :method_with=
+
+        def ==(other)
+          super
+        end
+        airbrake_capture_timing :==
+
+        def method_with_block
+          yield(1)
+        end
+        airbrake_capture_timing :method_with_block
+
+        def method_with_everything!(a, b = nil, *args, foo:, bar: nil)
+          raise ArgumentError if !a || !b || !foo || !bar || args.empty?
+          yield(1)
+        end
+        airbrake_capture_timing :method_with_everything!
+
+        def writer_with_everything=(a, b = nil, *args, foo:, bar: nil)
+          raise ArgumentError if !a || !b || !foo || !bar || args.empty?
+          yield(1)
+        end
+        airbrake_capture_timing :writer_with_everything=
+
+        prepend(
+          Module.new do
+            def prepended_method!(*args, **kw_args)
+              super
+            end
+
+            def prepended_writer=(*args, **kw_args)
+              super
+            end
+
+            protected
+
+            def a_prepended_protected_method
+              super
+            end
+
+            private
+
+            def a_prepended_private_method
+              super
+            end
+          end,
+        )
+
+        def prepended_method!(a, b = nil, *args, foo:, bar: nil)
+          raise ArgumentError if !a || !b || !foo || !bar || args.empty?
+          yield(1)
+        end
+        airbrake_capture_timing :prepended_method!
+
+        def prepended_writer=(a, b = nil, *args, foo:, bar: nil)
+          raise ArgumentError if !a || !b || !foo || !bar || args.empty?
+          yield(1)
+        end
+        airbrake_capture_timing :prepended_writer=
+
+        protected
+
+        def a_protected_method; end
+        airbrake_capture_timing :a_protected_method
+
+        def a_prepended_protected_method; end
+        airbrake_capture_timing :a_prepended_protected_method
+
+        private
+
+        def a_private_method; end
+        airbrake_capture_timing :a_private_method
+
+        def a_prepended_private_method; end
+        airbrake_capture_timing :a_prepended_private_method
       end
+    end
+
+    if defined?(Warning) && Warning.respond_to?(:warn)
+      # Make sure there are no Ruby 2.7+ warnings.
+      before do
+        expect(Warning).to_not receive(:warn).with(any_args)
+      end
+    end
+
+    it "doesn't generate any methods with invalid names" do
+      expect(klass.instance_methods + klass.private_instance_methods)
+        .to all match(/\A(\w+[?!=]?|\W+)\z/)
+    end
+
+    it "doesn't change visibility of public methods" do
+      expect(klass.public_instance_methods)
+        .to include(:method_with!, :prepended_method!)
+    end
+
+    it "doesn't change visibility of protected methods" do
+      expect(klass.protected_instance_methods)
+        .to include(:a_protected_method, :a_prepended_protected_method)
+    end
+
+    it "doesn't change visibility of private methods" do
+      expect(klass.private_instance_methods)
+        .to include(:a_private_method, :a_prepended_private_method)
+    end
+
+    it "returns the method name so other methods can use it" do
+      expect(klass.airbrake_capture_timing(:foo)).to be :foo
+    end
+
+    it "doesn't generate too many anonymous modules" do
+      # We should have two anonymous modules, including the one defined above.
+      expect(klass.ancestors.index(klass)).to eq 2
+      mod = klass.__send__("__airbrake_capture_timing_module__")
+      expect(klass.ancestors.index(mod)).to be < 2
     end
 
     context "when request store doesn't have any routes" do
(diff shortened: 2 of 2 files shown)


BrianHawley:
Will fail until airbrake/airbrake-ruby#542 is released.

BrianHawley:
The ruby-2.2 and ruby-2.7 errors are fixed by #1064.

Looking into the jruby bug, which is definitely a bug in jruby itself, but it's strictly an incompatibility with part of rspec. I'm going to see if I can rephrase the tests so rspec doesn't use the incompatible features.

kyrylo:
I gave this a shot and managed to slap together some code that passes most of the tests. Basically, the idea is to avoid the [`default_wrapping_style`](https://github.com/airbrake/airbrake-ruby/pull/542) option. I am not at all a fan of it because it clutters our API with unnecessary details, so I tried to find other ways to detect prepended methods automatically. I used your implementation as a reference (the Module.new trick).

So in the code below we detect prepended methods by checking the position of the method we "caputure" in the array of ancestors. We know for sure (I hope so) that the index of a prepended module is always lower than the index of the class that we prepend to. Included modules will have greater index.

```diff
diff --git a/lib/airbrake/rack/instrumentable.rb b/lib/airbrake/rack/instrumentable.rb
index 4776274..9f21836 100644
--- a/lib/airbrake/rack/instrumentable.rb
+++ b/lib/airbrake/rack/instrumentable.rb
@@ -14,12 +14,27 @@ module Airbrake
     # @api public
     # @since v9.2.0
     module Instrumentable
-      def airbrake_capture_timing(method_name, label: method_name.to_s)
+      def airbrake_capture_timing(method_name, label: method_name.to_s, with: nil)
+        self_class_idx = ancestors.index(self)
+        method_owner_idx = ancestors.index(instance_method(method_name).owner)
         alias_method "#{method_name}_without_airbrake", method_name

-        define_method(method_name) do |*args|
-          Airbrake::Rack.capture_timing(label) do
-            __send__("#{method_name}_without_airbrake", *args)
+
+        if method_owner_idx < self_class_idx
+          prepend(
+            Module.new do
+              define_method(method_name) do |*args, &block|
+                Airbrake::Rack.capture_timing(label) do
+                  __send__("#{method_name}_without_airbrake", *args, &block)
+                end
+              end
+            end
+          )
+        else
+          define_method(method_name) do |*args, &block|
+            Airbrake::Rack.capture_timing(label) do
+              __send__("#{method_name}_without_airbrake", *args, &block)
+            end
           end
         end
       end
```

What do you think?

BrianHawley:
I think that the tests I wrote are sufficient to check whether that works without stack overflows. It helps that we've narrowed it down to the prepend happening before the chain. I checked, and prepend happening after the chain works. We'll still need to do the Ruby 2.7 argument forwarding check, but we can do that inside the method if needed, or even define the constant in this file.

BrianHawley:
Removed the stack overflow tests, as they're no longer necessary. Added tests for method visibility and making sure that no invalid method names are generated.

BrianHawley:
New unrelated JRuby gem problem today. I'll see what I can do.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -14,15 +16,62 @@ module Rack
...
     # @since v9.2.0
     module Instrumentable
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/PerceivedComplexity
       def airbrake_capture_timing(method_name, label: method_name.to_s)
-        alias_method "#{method_name}_without_airbrake", method_name
+        # Check the visibility.
+        is_public = private_method_defined?(method_name)
```
kyrylo: Good idea to preserve visibility, I didn't think of this one 👍

lib/airbrake/rack/instrumentable.rb
```diff
@@ -14,15 +16,62 @@ module Rack
...
-        define_method(method_name) do |*args|
-          Airbrake::Rack.capture_timing(label) do
-            __send__("#{method_name}_without_airbrake", *args)
-          end
+        # Don't chain already-prepended or operator methods.
+        self_class_idx = ancestors.index(self)
+        method_owner_idx = ancestors.index(instance_method(method_name).owner)
+        if method_owner_idx < self_class_idx || /\A\W/ =~ method_name
+          prepend(
+            Module.new do
```
kyrylo: While this trick is useful, it has issues. I ran the test suite and inspected ancestors of the test class that you defined. As you can see, we prepend a module every time we call `airbrake_capture_timing`. This is expected but makes debugging harder now.

```
> klass.ancestors
=> [#<Module:0x00007f8875a011a8>,
 #<Module:0x00007f8875a03750>,
 #<Module:0x00007f8875bf8948>,
 #<Module:0x00007f8875bf8ec0>,
 #<Module:0x00007f8875bf90a0>,
 #<Module:0x00007f8875bfa900>,
 #<Class:0x00007f88751d51a0>,
 Object,
 JSON::Ext::Generator::GeneratorMethods::Object,
 PP::ObjectMixin,
 Kernel,
 BasicObject]
```

I think for this PR anonymous modules will do. In the future (prior to releasing the library), we would need to solve this to improve debugging. We would probably need to wrap `Module` in our own class, so that we could control naming. Another idea is to prepend only to one module and if we are prepending again to the same class, we would actually define a method on our already prepended module, so that we don't waste memory by spawning modules with 1 method only. Either way, let's not do anything clever in this PR.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -14,15 +16,62 @@ module Rack
...
+                  end
+                end
+              RUBY
+              # Restore the visibility.
+              protected method_name if is_protected
+              private method_name if is_private
+            end,
+          )
+        else
+          aliased = method_name.to_s.sub(/([?!=])$/, '')
```
kyrylo: What's the purpose of handling punctuation? If this is just a cosmetic change, I would prefer not to mess with it.

`method_with!_without_airbrake` is clear enough because the `without_` pattern is very common in the Ruby world. By juggling punctuation we might only confuse people (even though it looks nicer).

kyrylo:
FWIW you may want to rebase this on `master` since I released airbrake-ruby with your fixes.

spec/unit/rack/instrumentable_spec.rb
```diff
@@ -25,9 +26,108 @@ def method_with_args_and_vla(*args); end
...
+        def a_private_method; end
+        airbrake_capture_timing :a_private_method
+
+        def a_prepended_private_method; end
+        airbrake_capture_timing :a_prepended_private_method
+      end
+      # rubocop:enable Style/SingleLineMethods, Lint/UnusedMethodArgument
+    end
+
+    if defined?(Warning) && Warning.respond_to?(:warn)
```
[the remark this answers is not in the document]
BrianHawley: This code is verifying that it doesn't warn for deprecated behavior. If you undo the Ruby 2.7 fix in the method, this test definitely fails. This before block doesn't fix warnings it the tests; it verifies that there are no warnings generated by the tests.

spec/unit/rack/instrumentable_spec.rb
```diff
@@ -5,6 +5,7 @@
     let(:routes) { Airbrake::Rack::RequestStore[:routes] }
 
     let(:klass) do
+      # rubocop:disable Style/SingleLineMethods, Lint/UnusedMethodArgument
```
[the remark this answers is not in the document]
BrianHawley: Sounds good to me.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -14,15 +16,62 @@ module Rack
     # @api public
     # @since v9.2.0
     module Instrumentable
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/MethodLength
+      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/PerceivedComplexity
```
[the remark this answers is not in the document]
BrianHawley: The only reason this code is so complex is because it needs to be to handle the various problems that come up in this situation. I can move the code to self methods on the module so they don't pollute the namespace of the classes that include this module, and that will remove the warnings, but what I can't do is make it less complex for real.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -14,15 +16,62 @@ module Rack
...
+              end
+            end
+          RUBY
+          # Restore the visibility.
+          public method_name if is_public
+          protected method_name if is_protected
+          private method_name if is_private
         end
+
+        method_name # for chaining
```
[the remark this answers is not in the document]
BrianHawley: This method isn't meant to be used by *your* Ruby code. This is a public facing API to be used in other people's Ruby code, and some of those other people do write their code in that way. It costs nothing to set a useful return value.

lib/airbrake/rack/instrumentable.rb
BrianHawley: It's also not uncommon when using chained methods to call the wrapped methods outside of the wrapper. Not fixing the punctuation just makes that kind of thing more difficult. Also, I don't want to generate methods that break syntax when future Ruby versions might start checking that kind of thing, since it's an obvious feature that I've seen added to other languages.

lib/airbrake/rack/instrumentable.rb
BrianHawley: I agree with your assessment of these issues. I've had to apply this code pattern in other projects, some of which I'm working on right now, and this is consistent with what I've seen before. It would be worth it to me to work on a better solution, as I might want to apply that solution elsewhere. Your suggestions are appreciated.

Nonetheless, I also agree that there's no need to have Ruby 2.7 compatibility wait for this. Best to get a working solution out now, and clean it up later. If I can't get a better solution out in the next round of fixes, good, but if not it should be OK.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -14,15 +16,62 @@ module Rack
...
+      # rubocop:disable Metrics/PerceivedComplexity
       def airbrake_capture_timing(method_name, label: method_name.to_s)
-        alias_method "#{method_name}_without_airbrake", method_name
+        # Check the visibility.
+        is_public = private_method_defined?(method_name)
+        is_protected = protected_method_defined?(method_name)
+        is_private = private_method_defined?(method_name)
+
+        # Make sure to forward keyword arguments too when using 2.7+ syntax.
+        use_kw_args = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.7")
```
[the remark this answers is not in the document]
BrianHawley: Only if we can avoid the constant being inherited by the classes and modules that are extending themselves with this one. I'll think of a solution.

BrianHawley:
Fixed the multiple modules issue too, and added a test for it. Still anonymous, but cached.

spec/unit/rack/instrumentable_spec.rb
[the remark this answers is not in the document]
BrianHawley: Btw, those warnings in the pry gems were legitimate bugs in pry. I use `binding.irb` now.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+        if mod.should_prepend?(self, method_name)
+          mod.prepend_airbrake_capture_timing(self, method_name, label)
+        else
+          mod.chain_airbrake_capture_timing(self, method_name, label)
+        end
+        method_name
+      end
+
+      # @api private
+      def __airbrake_capture_timing_module__
```
kyrylo: Even though this is marked as private API, the method is still public and therefore extenders of the module will have access to it (for no good reason). Perhaps you left it private for easier testability but I would prefer not to leak implementation details to user space. We can work around this in the tests by mocks or calling `__send__`.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+
+          # Restore the visibility.
+          public method_name if is_public
+          protected method_name if is_protected
+          private method_name if is_private
+        end
+      end
+
+      # @api private
+      # Using a method instead of a constant so it isn't defined in the target.
```
kyrylo: If we use a constant, it won't be defined in the target. That said, it indeed will be available from the target's context, true.

```rb
[1] pry(main)> module M
[1] pry(main)*   FOO = 1
[1] pry(main)* end
=> 1
[2] pry(main)> class A
[2] pry(main)*   extend M
[2] pry(main)* end
=> A
[3] pry(main)> class A
[3] pry(main)*   class << self
[3] pry(main)*     p FOO
[3] pry(main)*   end
[3] pry(main)* end
1
=> 1
```

Therefore, I think we should say this:

```suggestion
      # Using a method instead of a constant so it isn't accessible in the target.
```

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+          prepend mod
+        end
+      end
+
+      # @api private
+      def self.chain_airbrake_capture_timing(klass, method_name, label)
+        args = method_signature
+        klass.module_exec do
+          # Check the visibility.
+          is_public = private_method_defined?(method_name)
```
kyrylo:
```suggestion
          is_public = public_method_defined?(method_name)
```

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+      end
 
-        define_method(method_name) do |*args|
-          Airbrake::Rack.capture_timing(label) do
-            __send__("#{method_name}_without_airbrake", *args)
+      # @api private
+      def self.prepend_airbrake_capture_timing(klass, method_name, label)
+        args = method_signature
+        klass.module_exec do
+          # Check the visibility.
```
kyrylo: I don't like how we repeat this code for both cases so I gave this a shot and factored out it into a separate method. 
* No need to define 3 temp variables, just 1 now
* We declare visibility directly when evalling

<details>
<summary>click to see the diff</summary>

```diff
diff --git a/lib/airbrake/rack/instrumentable.rb b/lib/airbrake/rack/instrumentable.rb
index 8d6b349..da0c702 100644
--- a/lib/airbrake/rack/instrumentable.rb
+++ b/lib/airbrake/rack/instrumentable.rb
@@ -50,10 +50,9 @@ module Airbrake
       # @api private
       def self.prepend_airbrake_capture_timing(klass, method_name, label)
         args = method_signature
+        visibility = check_visibility(klass, method_name)
+
         klass.module_exec do
-          # Check the visibility.
-          is_protected = protected_method_defined?(method_name)
-          is_private = private_method_defined?(method_name)
           mod = __airbrake_capture_timing_module__
           mod.module_exec do
             # Generate the wrapper method.
@@ -63,10 +62,8 @@ module Airbrake
                   super
                 end
               end
+              #{visibility} :#{method_name}
             RUBY
-            # Restore the visibility.
-            protected method_name if is_protected
-            private method_name if is_private
           end
           prepend mod
         end
@@ -75,29 +72,37 @@ module Airbrake
       # @api private
       def self.chain_airbrake_capture_timing(klass, method_name, label)
         args = method_signature
-        klass.module_exec do
-          # Check the visibility.
-          is_public = private_method_defined?(method_name)
-          is_protected = protected_method_defined?(method_name)
-          is_private = private_method_defined?(method_name)
+        visibility = check_visibility(klass, method_name)
+
+        # Generate the wrapper method.
+        aliased = method_name.to_s.sub(/([?!=])$/, '')
+        punctuation = Regexp.last_match(1)
+        wrapped_method_name = "#{aliased}_without_airbrake#{punctuation}"

-          # Generate the wrapper method.
-          aliased = method_name.to_s.sub(/([?!=])$/, '')
-          punctuation = Regexp.last_match(1)
-          alias_method "#{aliased}_without_airbrake#{punctuation}", method_name
+        klass.module_exec do
+          alias_method wrapped_method_name, method_name
           remove_method method_name
           module_eval <<-RUBY, __FILE__, __LINE__ + 1
             def #{method_name}(#{args})
               Airbrake::Rack.capture_timing(#{label.to_s.inspect}) do
-                __send__("#{aliased}_without_airbrake#{punctuation}", #{args})
+                __send__(:#{wrapped_method_name}, #{args})
               end
             end
+            #{visibility} :#{method_name}
           RUBY
+        end
+      end

-          # Restore the visibility.
-          public method_name if is_public
-          protected method_name if is_protected
-          private method_name if is_private
+      # @api private
+      def self.check_visibility(klass, method_name)
+        klass.module_exec do
+          if protected_method_defined?(method_name)
+            'protected'
+          elsif private_method_defined?(method_name)
+            'private'
+          else
+            'public'
+          end
         end
       end
</details>

File: lib/airbrake/rack/instrumentable.rb [hunk 13]
```diff
@@ -15,12 +17,99 @@ module Rack
     # @since v9.2.0
     module Instrumentable
       def airbrake_capture_timing(method_name, label: method_name.to_s)
-        alias_method "#{method_name}_without_airbrake", method_name
+        mod = ::Airbrake::Rack::Instrumentable
```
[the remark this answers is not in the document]
BrianHawley: Sounds good. I'll do that.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+        if mod.should_prepend?(self, method_name)
+          mod.prepend_airbrake_capture_timing(self, method_name, label)
+        else
+          mod.chain_airbrake_capture_timing(self, method_name, label)
+        end
+        method_name
```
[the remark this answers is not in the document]
BrianHawley: Sounds good. I'll add one.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+        klass.module_exec do
+          # Check the visibility.
+          is_protected = protected_method_defined?(method_name)
+          is_private = private_method_defined?(method_name)
+          mod = __airbrake_capture_timing_module__
+          mod.module_exec do
+            # Generate the wrapper method.
+            module_eval <<-RUBY, __FILE__, __LINE__ + 1
+              def #{method_name}(#{args})
+                Airbrake::Rack.capture_timing(#{label.to_s.inspect}) do
```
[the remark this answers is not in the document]
BrianHawley: We need to call `.inspect` because it adds the quotes around the string, because this is interpolating into source code. We call the `.to_s` because I don't want to make too many assumptions about what label is, and string inspect definitely generates source compatible values. On 2.7+, `.to_s.inspect` even returns the same frozen string every time for symbols. We can get rid of the `.to_s` safely for values of `label` that the code supports, if you like, but we need the `.inspect`.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+
+      # @api private
+      # Using a method instead of a constant so it isn't defined in the target.
+      if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.7")
+        def self.method_signature
+          "*args, **kw_args, &block"
+        end
+      else
+        def self.method_signature
+          "*args, &block"
```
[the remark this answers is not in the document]
BrianHawley: I'm OK with adding `.freeze` on it. I put the frozen string literal at the top of the file, which will avoid the string allocation on Ruby 2.3+, but `.freeze` on a literal string is resolved at compile time on 2.3 and some earlier versions so it wouldn't hurt to add here.

lib/airbrake/rack/instrumentable.rb
BrianHawley: Sounds good to me.

lib/airbrake/rack/instrumentable.rb
BrianHawley: Whoops :)

lib/airbrake/rack/instrumentable.rb
BrianHawley: I like it.

lib/airbrake/rack/instrumentable.rb
BrianHawley: We're leaking an instance variable in any case. The method only exists at all for testing purposes, and in case we have to do something weird for thread safety in the future, though I made sure that it would be thread-safe enough for our purposes by saving the result in a local variable at the call site.

There is no harm in making the method available; it's only api private because it's useless to the user, so there's no point in documenting it.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,99 @@ module Rack
...
+          module_eval <<-RUBY, __FILE__, __LINE__ + 1
+            def #{method_name}(#{args})
+              Airbrake::Rack.capture_timing(#{label.to_s.inspect}) do
+                __send__("#{aliased}_without_airbrake#{punctuation}", #{args})
+              end
+            end
+          RUBY
+
+          # Restore the visibility.
+          public method_name if is_public
```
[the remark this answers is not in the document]
BrianHawley: If we don't have this check in the chain method, something like this would mess with the visibility:
```
def foo; end
private
airbrake_capture_timing :foo
```

lib/airbrake/rack/instrumentable.rb
BrianHawley: Sorry, I tried to remove the method and just use the instance variable, and I couldn't make it thread-safe and still work without a lot of effort. I'll make the method private so it doesn't add a public method in controllers.

lib/airbrake/rack/instrumentable.rb
```diff
@@ -15,12 +17,109 @@ module Rack
...
+        # Generate the wrapper method.
+        aliased = method_name.to_s.sub(/([?!=])$/, '')
+        punctuation = Regexp.last_match(1)
+        wrapped_method_name = "#{aliased}_without_airbrake#{punctuation}"
+        klass.module_exec do
+          alias_method wrapped_method_name, method_name
+          # Avoid a redefinition warning.
+          begin
+            remove_method method_name
+          rescue NameError
```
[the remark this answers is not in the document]
BrianHawley: This is super annoying. The only efficient alternative to this requires Ruby 2.6+ and its second inherit parameter to `method_defined?` and `private_method_defined?`. There are so many problems with chaining style.

kyrylo:
Thanks for the fixes and your patience. This is looking good to merge now :+1:

PR #1070, opened 2020-01-29
Make sure there are frozen string literals

BrianHawley:
Turns out that they were added in Ruby 2.3, not 3.0.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,9 +30,13 @@ Style/NumericLiterals:
 Style/SignalException:
   EnforcedStyle: only_raise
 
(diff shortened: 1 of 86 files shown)


kyrylo:
FWIW, according to https://bugs.ruby-lang.org/issues/11473#note-53 Ruby 3.0 will not have frozen string literals by default.

BrianHawley:
@kyrylo I am more concerned with the frozen string literal setting comments. It tends to improve performance a bit and reduce overhead if you turn that setting on. Having it not be the default makes that even more important.

PR #1110, opened 2020-07-15
Upload private source maps via Capistrano

danieldimitrov:
Hello Airbrake team!

I've recently migrated one project I am working on to Airbrake and I would like to propose one new addition to the gem.

### Motivation
We use in our projects a bunch of Capistrano after:deploy hooks so it is convenient to be able to upload private source maps via another Capistrano hook instead of, for example, some Webpack plugin. That way, we could have all our "after deploy" processes at one place and organized well, instead of having them spread in different tools.

### Problem
After reading Airbrake documentation and internals of this gem I didn't found any native support for this. So why not add it?

### Example of usage

```ruby
#Capfile
...

require "airbrake/capistrano"

set :airbrake_project_id, ENV["AIRBRAKE_PROJECT_ID"]
set :airbrake_project_key, ENV["AIRBRAKE_PROJECT_KEY"]
set :airbrake_sourcemaps_target_dir, "public/packs/js"

...
after "deploy", "airbrake:upload_private_sourcemaps"
```

```ruby
#config/deploy/staging.rb or other environemnt-related deploy script file

set :airbrake_sourcemaps_minified_url_base, "https://YOUR_URL.com"
```

### The final disclaimer
The code in this PR was written just to demonstrate the idea. I expect that it would need much more love before the final merge. I would love to continue working on it and finish it if the idea gets enough support.


kyrylo:
I am not against the idea for this kind of support but I do have a couple thoughts/questions.

The code you wrote doesn't seem to be dependent on Capistrano in any way. Can we make it a separate class so that it can be used in any environment? Then Capistrano could hook instantiate the class and provide the params it needs. The code for that class may possibly live in https://github.com/airbrake/airbrake-ruby

I'm not familiar with `multipart-post` but I am fairly confident we don't want to introduce this dependency for everybody who uses this gem because the majority of them are not Capistrano users. Do we really need it?

danieldimitrov:
Good ideas! I think both are valid. Tbh, I also don't like the idea of a new dependency, and I don't think we need it.

Regarding the code, I think that only Capistrano-related part is the task. So if I would move logic to airbrake-ruby, here we could have something like:

```ruby
namespace :airbrake do
   desc "Upload private sourcemaps to Airbrake"
   task :upload_private_sourcemaps do
     on roles fetch(:rollbar_role) do
       SourcemapUploader.new(params).upload
     end
   end
end
```

Does it make sense?

kyrylo:
Looking good. I would modify the public API a bit so it looks more consistent with the rest of the library:

```
sourcemap = Airbrake::Sourcemap.new(params)
Airbrake.upload_sourcemap(sourcemap)
```

kyrylo:
@danieldimitrov just to clarify, are you willing to work on this or is it okay for me to continue from here?

danieldimitrov:
@kyrylo I would like to work on it. I am going to submit PRs during the next week.

kyrylo:
Nice one, thanks

PR #1111, opened 2020-07-17
rails/app: fix request.path with "//"

### LINKED ISSUE #1072: request.path starts with "//" after updating to v10
I've updated airbrake to avoid deprecation warnings
https://github.com/airbrake/airbrake-ruby/issues/532

when opening a page in active admin, `request.path` is now
`//admin/users/1770929`
before it was
`/admin/users/1770929`

The only change in the gemfile lock is concerning airbrake

before
```
    airbrake (9.3.0)
      airbrake-ruby (~> 4.5)
    airbrake-ruby (4.13.0)
```

after
```
    airbrake (10.0.1)
      airbrake-ruby (~> 4.13)
    airbrake-ruby (4.13.1)
```

edit : a bit more details
- its a rails 3.2.22.5 app with an engine "backoffice"
- it happens regardless how many starting slashes I enter in the URL in my browser

kyrylo:
Fixes #1072 (request.path starts with "//" after updating to v10)

This is a hack but we have no better option since `request.dup` is not able to
duplicate `script_name` properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Changelog
 
 ### master
 
+* Rails APM: fixed double slash in front of a route name when mounting engines
+  at `/` ([#1111](https://github.com/airbrake/airbrake/pull/1111))
+
 ### [v10.1.0.rc.1][v10.1.0.rc.1] (July 14, 2020)
 
 * Fixes `Airbrake::Sidekiq::RetryableJobsFilter` erroneously reporting retry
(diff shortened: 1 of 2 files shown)


MathieuDerelle:
it is possible to release this in a `v10.0.6` ? 

first release of this is in v11 where you dropped support for ruby 2.2 (and my app is in ruby 2.2)

kyrylo:
Sure, just released v10.0.6: https://rubygems.org/gems/airbrake/versions/10.0.6

MathieuDerelle:
thanks 👍

PR #1202, opened 2021-11-14
Fix up the PR number in PR 1179

tnir:
The linked in CHANGELOG updated in #1179 was wrong.

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>


tnir:
cc @kyrylo

kyrylo:
Good catch, thanks!

Issue #1177, opened 2021-09-19
airbrake gem  don't work  with 7.0.0.alpha2 release


equivalent:
I'm trying out Rails 7.0.0.alpha2 on Heroku with fresh Rails project. All works well so I've decided to install Airbrake to capture any future errors. As soon as I've installed Airbrake gem prod environment start to crush once deployed to Heroku with: 

```
2021-09-19T19:44:56.893423+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] Started GET "/" for 141.101.96.9 at 2021-09-19 19:44:56 +0000
2021-09-19T19:44:56.895831+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] Processing by PostsController#index as HTML
2021-09-19T19:44:56.901553+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79]
2021-09-19T19:44:56.901554+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] NoMethodError (undefined method `glob?' for nil:NilClass):
2021-09-19T19:44:56.901554+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79]
2021-09-19T19:44:56.901555+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] actionpack (7.0.0.alpha2) lib/action_dispatch/journey/route.rb:138:in `glob?'
2021-09-19T19:44:56.901556+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] airbrake (11.0.3) lib/airbrake/rails/app.rb:48:in `block (2 levels) in recognize_route'
2021-09-19T19:44:56.901557+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] actionpack (7.0.0.alpha2) lib/action_dispatch/journey/router.rb:74:in `block in recognize'
2021-09-19T19:44:56.901557+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] actionpack (7.0.0.alpha2) lib/action_dispatch/journey/router.rb:66:in `each'
2021-09-19T19:44:56.901558+00:00 app[web.1]: [37244e36-bd79-486d-894e-5d3dddb07d79] actionpack (7.0.0.alpha2) lib/action_dispatch/journey/router.rb:66:in `recognize'
[...]

Issue #1091, opened 2020-05-20
Performance - all routes appear as / with 10.0.3. Works fine in 9.5.5 


thekendalmiller:
I think this is the same issue as https://github.com/airbrake/airbrake/issues/1072

On the airbrake performance dashboard, all of the performance requests are appearing as /*path instead of the actual path. This is the page I'm talking about...removed my project_id airbrake.io/projects/#{project_id}/performance?

If I downgrade to 9.5.5, it works like I would expect and stores the request path as admin/users/show

thekendalmiller:
Also reading the previous issue, the app I'm on also has an engine. 
I removed the routes.rb mount for the engine but it still isn't working. 
I then removed all of the routes the engine adds so that the the engine routes look like this
```
Cms::Engine.routes.draw do
 # stuff
 # commented
 # out
end
```

And airbrake works.. clearly not acceptable because I need the routes from the engine.
Just debugging and thought I'd share what I had notice in case you could point me in a direction.

This app is rails 4.2.x

thekendalmiller:
To narrow it down further, it's one specific route in the engine that seems to break this - a catch all route. Airbrake Performance reporting works correctly as long as the engine doesn't have a catch all in it. 

```
Cms::Engine.routes.draw do
 # a
 # bunch
 # of 
 # routes
 get '*path' => 'static_contents#show', param: :path
end
```

destari:
Thanks! We'll take a look!

thekendalmiller:
Thanks. I was curious and spent a few more minutes.
This is the commit that introduced the issue: 6836d58d8579aaa4f367ae7c332b37e0c0b4a643
From: https://github.com/airbrake/airbrake/pull/1059

My guess is this line https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rails/app.rb#L28 is the issue. It probably needs to make sure the engine doesn't recognize the route using a catchall.

thekendalmiller:
Also, I peaked at how other APM gems get the controller and path form rails.
They are using: https://guides.rubyonrails.org/active_support_instrumentation.html#process-action-action-controller

Maybe you could get the controller and path in https://github.com/airbrake/airbrake/blob/master/lib/airbrake/rails/event.rb

kyrylo:
@thekendalmiller thanks for the detailed description. That was really useful. I released v10.0.4 with the fix.

thekendalmiller:
[...]

PR #1206, opened 2021-12-26
Workaround for route.glob? on Rails 7.0.0

tnir:
Workaround for route.glob? on Rails 7.0.0 to address #1177.

Updates #1092 (#1091)

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

### DIFF
--- a/lib/airbrake/rails/app.rb
+++ b/lib/airbrake/rails/app.rb
@@ -39,13 +39,7 @@ module Airbrake
 
             # Skip "catch-all" routes such as:
             #   get '*path => 'pages#about'
-            #
-            # @todo The `glob?` method was added in Rails v4.2.0.beta1. We
-            # should remove the `respond_to?` check once we drop old Rails
-            # versions support.
-            #
-            # https://github.com/rails/rails/commit/5460591f0226a9d248b7b4f89186bd5553e7768f
-            next if route.respond_to?(:glob?) && route.glob?
+            next if route.path.ast.nil? && route.glob?
 
             path =
               if engine == ::Rails.application


tnir:
@kyrylo Comments?

kyrylo:
@tnir thanks for the good work. I am currently busy with something else, I will try to schedule some time for Rails 7 support early next year.

tnir:
@kyrylo Could you mind take a look at this?

kyrylo:
@tnir I'll take a look next week as I am on vacation at the moment.

lib/airbrake/rails/app.rb
```diff
@@ -45,7 +45,7 @@ def self.recognize_route(request)
             # versions support.
             #
             # https://github.com/rails/rails/commit/5460591f0226a9d248b7b4f89186bd5553e7768f
-            next if route.respond_to?(:glob?) && route.glob?
+            next if route.nil? && route.respond_to?(:glob?) && route.glob?
```
kyrylo, line 48: I think we can safely remove `route.respond_to?(:glob?)` (and the todo comment above) since we no longer test against Rails 4.1 and lower.

Can `route` actually be ever `nil`? In the failing scenario it's not `nil`. What is `nil` is `route.path.ast`. Should we be checking for `route.path.ast.nil?` since that's the implementation of `glob?` and the call that fails?

tnir, line 48: So does the following change with removing `@todo` section above make sense?
```suggestion
            next if route.path.ast.nil? && route.glob?
```

tnir: https://github.com/airbrake/airbrake/pull/1206#discussion_r782924573

lib/airbrake/rails/app.rb
```diff
@@ -39,13 +39,7 @@ def self.recognize_route(request)
...
             # Skip "catch-all" routes such as:
             #   get '*path => 'pages#about'
-            #
-            # @todo The `glob?` method was added in Rails v4.2.0.beta1. We
-            # should remove the `respond_to?` check once we drop old Rails
-            # versions support.
-            #
-            # https://github.com/rails/rails/commit/5460591f0226a9d248b7b4f89186bd5553e7768f
-            next if route.respond_to?(:glob?) && route.glob?
+            next if route.path.ast.nil? && route.glob?
```
klausmeyer, line 42: With this version I'm getting the following error in my rails 7.0.1 apps (`with CI=true` eager loading the code):

```
  9) GET /api/v1/user without a valid api token returns with status code 401
     Failure/Error: get '/api/v1/user'

     NoMethodError:
       undefined method `glob?' for nil:NilClass

               path.ast.glob?
                       ^^^^^^
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/bundler/gems/airbrake-601193db2c93/lib/airbrake/rails/app.rb:42:in `block (2 levels) in recognize_route'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/bundler/gems/airbrake-601193db2c93/lib/airbrake/rails/app.rb:35:in `block in recognize_route'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/bundler/gems/airbrake-601193db2c93/lib/airbrake/rails/app.rb:34:in `each'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/bundler/gems/airbrake-601193db2c93/lib/airbrake/rails/app.rb:34:in `recognize_route'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/bundler/gems/airbrake-601193db2c93/lib/airbrake/rails/action_controller_route_subscriber.rb:20:in `call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/warden-1.2.9/lib/warden/manager.rb:36:in `block in call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/warden-1.2.9/lib/warden/manager.rb:34:in `catch'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/warden-1.2.9/lib/warden/manager.rb:34:in `call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/rack-2.2.3/lib/rack/tempfile_reaper.rb:15:in `call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/rack-2.2.3/lib/rack/etag.rb:27:in `call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/rack-2.2.3/lib/rack/conditional_get.rb:27:in `call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/rack-2.2.3/lib/rack/head.rb:12:in `call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/rack-2.2.3/lib/rack/session/abstract/id.rb:266:in `context'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/gems/rack-2.2.3/lib/rack/session/abstract/id.rb:260:in `call'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/bundler/gems/airbrake-601193db2c93/lib/airbrake/rack/middleware.rb:34:in `call!'
     # /Users/meyer/.rvm/gems/ruby-3.1.0@whatchy/bundler/gems/airbrake-601193db2c93/lib/airbrake/rack/middleware.rb:23:in `call'
```

The error is actually coming from this [LOC](https://github.com/rails/rails/blob/9fb6cdf8fafea8d23fae06586f0f3248f873d53c/actionpack/lib/action_dispatch/journey/route.rb#L138) in rails:

```ruby
def glob?
  path.ast.glob?
end
```

Shouldn't the check therefore be like this instead?

```diff
- next if route.path.ast.nil? && route.glob?
+ next if !route.path.ast.nil? && route.glob?
```

Perhaps it would be a good idea to add a "catch-all" to the [routes in the dummy rails app](https://github.com/airbrake/airbrake/blob/4d7cacb7c3cef6b3d7f4cecb31bc16afeee7b832/spec/apps/rails/dummy_app.rb#L40-L60)?

Let me know if I can help somehow.

kyrylo: I forgot to submit my review. It should be

```suggestion
            next if route.path.ast.nil? || route.glob?
```

Either way, I had a chance to speak to someone who was modifying the code in Rails and we need a different approach here. I will submit a PR and release a new version. @klausmeyer would you be able to test my change against your app prior to releasing a new version?

klausmeyer:
> Either way, I had a chance to speak to someone who was modifying the code in Rails and we need a different approach here. I will submit a PR and release a new version. @klausmeyer would you be able to test my change against your app prior to releasing a new version?

Yes sure. Just ping me and I'll give it a try.

lib/airbrake/rails/app.rb
klausmeyer: Yes sure. Just ping me and I'll give it a try.

kyrylo: Thank you very much @klausmeyer. Here's the PR: https://github.com/airbrake/airbrake/pull/1211
Branch: `1177-rails-7-glob-fix`

klausmeyer: 👀 I'm on it

klausmeyer: ✅ RSpec Suite of my 7.0.1 app with `config.eager_load = true`
✅ Deployment to Heroku
✅ Error reporting with `rake airbrake:test`

Used versions:

```
$ grep -A2 airbrake Gemfile.lock
  remote: https://github.com/airbrake/airbrake.git
  revision: 7e03129eef591c46cce1c465533c3efcab03d09e
  branch: 1177-rails-7-glob-fix
--
    airbrake (12.0.0)
      airbrake-ruby (~> 6.0)

GIT
--
    airbrake-ruby (6.0.2)
      rbtree3 (~> 0.5)
    autoprefixer-rails (10.3.3.0)
--
  airbrake!
  bootsnap
  bootstrap (~> 4.3)
```

kyrylo: Thanks a bunch!

kyrylo:
Closed in favor of https://github.com/airbrake/airbrake/pull/1211.

PR #1211, opened 2022-01-17
rails/app: check for `Nodes::Star` instead of using `glob?`

kyrylo:
Fixes https://github.com/airbrake/airbrake/issues/1177
(airbrake gem don't work with 7.0.0.alpha2 release)

Alternative to https://github.com/airbrake/airbrake/pull/1206

`glob?` on Rails 7+ raises error because Rails 7+ doesn't keep the route AST in
memory anymore. A Rails contributor who worked on the related code
[suggested][1] to use `Nodes::Star` instead, since that used to be the previous
implementation of the `glob?` method.

[1]: https://github.com/rails/rails/pull/43006#discussion_r783840849

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ Breaking changes:
 * Dropped support for Ruby 2.5
   ([#1208](https://github.com/airbrake/airbrake/issues/1208))
 
+Bug fixes:
+
+* Fixed APM not working on Rails 7 due to ``NoMethodError (undefined method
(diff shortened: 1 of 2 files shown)


klausmeyer (approved):
🚀 https://github.com/airbrake/airbrake/pull/1206#discussion_r786217054

gdott9:
I tested this PR in my app and it's working for me too. :+1:

tnir:
Thanks!

Issue #1239, opened 2022-10-07
Premature notification in Airbrake::Sidekiq::RetryableJobsFilter


dgsangoma:
We've set a global Sidekiq default of 5 for `max_retries`, but overridden that in the code for particular jobs to be higher like in the `ProgressWorker` job referenced below that has been increased via the option `sidekiq_options retry: 60`. In the example Parameters section below from an example Airbrake, you can see the following values that I think indicates the Airbrake notification shouldn't have been sent:

```
    "retry": 60,
    "retry_count": 5
```

Is it possible there's a bug in `RetryableJobsFilter` where it gets confused and defers to the global default `max_retries` instead of the per-job value?

In `config/initializers/airbrake.rb`, we have the `RetryableJobsFilter` enabled as follows:
```
...
require "airbrake/sidekiq"
...
Airbrake.add_filter(Airbrake::Sidekiq::RetryableJobsFilter.new)
```

Sample parameters:
```
{
  "job": {
    "args": [
      "969e17f3-e624-4630-abfe-6b614dd34db2",
      "2022-10-05T23:36:55+00:00",
      "2022-10-06T00:36:55+00:00",
      "wj0m3m5gbhozmpzt",
      "102617"
    ],
    "class": "ProgressWorker",
    "created_at": 1665013183.505255,
    "enqueued_at": 1665013645.800618,
    "error_class": "InProgressError",
    "error_message": "InProgressError",
    "failed_at": 1665013187.0334604,
    "jid": "fcf2ff69cde5365afdd0445a",
    "queue": "default",
    "retried_at": 1665013579.586878,
    "retry": 60,
    "retry_count": 5
  }
}
```

Environment:
```
{
  "program_name": "sidekiq 6.4.2 app [1 of 10 busy]"
}
```

Context:
```
{
  "action": "ProgressWorker",
  "component": "sidekiq",
  "environment": "production",
  "error_message": "InProgressError",
  "hostname": "bb694925539f",
  "language": "ruby/3.1.2",
  "messageParams": {
    "0": "InProgressError"
  },
  "messagePattern": "{}",
  "notifier": {
    "name": "airbrake-ruby",
    "url": "https://github.com/airbrake/airbrake-ruby",
    "version": "6.2.0"
  },
  "os": "x86_64-linux",
  "remoteAddr": "[REDACTED]",
  "remoteCountry": "United States",
  "remoteCountryCode": "US",
  "rootDirectory": "/usr/src/app",
  "severity": "error"
[...]

PR #1244, opened 2023-03-13
Separate var when overriding max retries before notify in Sidekiq jobs

dgsangoma:
Previously, the @max_retries var was being used for both the global default retry limit as well as when a developer indicates they want error notifications before retries have been exhausted for a Sidekiq job. This results in an undesirable side effect for apps that:

* Do not pass max_retries as an arg to the RetryableJobsFilter initializer to indicate they want early notifications for errors when there are retries remaining

* Explicitly set a higher retry count than the global default for some jobs

Since Sidekiq may reuse worker threads, this will result in a situation where a thread may first execute a job with no retry limit explicitly set, which will load the global default into @max_retries. Later that same thread will be used to execute a job that has an explicit retry limit set, but RetryableJobsFilter will mistakenly use the previously set value for @max_retries instead of first checking for the presence of the explicity retry limit on the job hash itself.

Fix this behavior by using a separate variable (@retries_before_notify) when a developer wants to enable early notifications when there are still retries remaining so it doesn't confuse

Hopefully fixes https://github.com/airbrake/airbrake/issues/1239

### DIFF
--- a/lib/airbrake/sidekiq/retryable_jobs_filter.rb
+++ b/lib/airbrake/sidekiq/retryable_jobs_filter.rb
@@ -16,7 +16,7 @@ module Airbrake
       end
 
       def initialize(max_retries: nil)
-        @max_retries = max_retries
+        @retries_before_notify = max_retries
       end
 
       def call(notice)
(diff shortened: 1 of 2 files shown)


dgsangoma:
@mmcdaris Is there any chance you might be able to bring this to the attention to someone at Airbrake who can review it?

mmcdaris (approved):
:+1: Thanks for the contribution, this will be included in the next release.

PR #1267, opened 2024-12-05
WIP - GH Actions

thompiler:
Testing branch for troubleshooting the GH actions failures


thompiler:
This check appears to be returning false for Jruby:
https://github.com/airbrake/airbrake-ruby/blob/7c9f1885717141ffe39f8a7900cea47b941be19b/airbrake-ruby.gemspec#L30-L34

So it's installing the wrong rbtree3

